PR #19, opened 2018-08-06
Projects YAML reference

mactsouk:


projects-yaml-reference_5b6813d20428631d7a89afa9.md
```diff
@@ -0,0 +1,117 @@
...
+
+Projects can also be added to Semaphore 2.0 with the `sem init` command. In
+that case, you will not need to create a YAML file on your own. However,
+`sem init` requires a local copy of the GitHub repository, even if that
+repository is not up to date, that does not contain a
+`.semaphore/semaphore.yml` file. Put simply, `sem init` is simpler but less
+flexible than the `sem create` command.
+
+Note that all branches of a GitHub repository that is added to Semaphore
+2.0 will be visible in Semaphore 2.0 automatically.
```
darkofabijan: This might be confusing. It will happen only when they make a push to branches.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
```diff
@@ -0,0 +1,117 @@
...
+
+## Properties
```
darkofabijan: If this is H2.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
```diff
@@ -0,0 +1,117 @@
...
+
+
+## apiVersion
```
darkofabijan: Then this should be H3

projects-yaml-reference_5b6813d20428631d7a89afa9.md
```diff
@@ -0,0 +1,117 @@
...
+
+List of value for `apiVersion`: `v1alpha`.
+
+## kind
+
+The value of the `kind` property is a string that specifies the type of the
+YAML file, that is whether it is for creating new projects or new `secrets`
+buckets. For projects the value of the `kind` property should be `Project`.
+
+List of values for `kind`: `Project`, `Secret`.
```
darkofabijan: Not sure if knowing about this option is helping here or just causing confusion here. In reality only `Project` can be used. It just happens to be the case that secrets have at the moment same structure of the file.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
```diff
@@ -0,0 +1,117 @@
...
+
+## metadata
+
+The `metadata` property is for grouping other properties. Currently, the only
+supported property is `name`.
+
+### name
+
+The `name` property defines the name of the Semaphore 2.0 project, as it will
```
darkofabijan: @shiroyasha we need help here. Are there any limitations on the name? e.g. no spaces, only ascii + 0-0 etc?

projects-yaml-reference_5b6813d20428631d7a89afa9.md
```diff
@@ -0,0 +1,117 @@
...
+
+The `spec` property is currently used for holding the `repository` property.
+
+### repository
+
+The `repository` property is used for holding the `url` property.
+
+#### url
+
+The `url` property is a string that specifies the URL of the GitHub repository
```
darkofabijan: Mention which URL format we are expecting here.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
mactsouk: Removed it.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
mactsouk: Yes.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
mactsouk: and yes!

projects-yaml-reference_5b6813d20428631d7a89afa9.md
mactsouk: Removed the entire line.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
mactsouk: I tried creating a S2 project with a space in the name and I got `error: http status 500 with message "" received from upstream`. The same when using `€` (euro sign) in the project name.
I will add that information in the reference page and will wait confirmation from @shiroyasha.

mactsouk: The value of the `name` property should be unique among all Semaphore 2.0 projects of the same organization and must only contain ASCII characters without spaces.

projects-yaml-reference_5b6813d20428631d7a89afa9.md
mactsouk: The format of the `url` value should be as follows: `git@github.com:github_username/github_repository.git`

PR #27, opened 2018-08-28
Addressing Damjan notes

mactsouk:


darkofabijan:
@mactsouk this is ok for merging once deps are removed.

PR #28, opened 2018-08-28
A documentation page for explaining Secrets for the average user.

mactsouk:


darkofabijan:
@mactsouk one general note here. Don't combine `secret` with `bucket`. Just say `secret`.

mactsouk:
@darkofabijan will do.

using-secrets_5b8506120428631d7a8a7239.md
```diff
@@ -387,17 +413,69 @@ file of that secret and execute `sem create`.
...
+	    os_image: ubuntu1804
+    
+	blocks:
+	  - name: Using SSH key files
+	    task:
+	      jobs:
+	      - name: Using server_key
+	        commands:
+	          - ls -l ~/.ssh
+			  - chmod 600 ~/.ssh/server_key
```
shiroyasha: By looking at this, I might even argue that you should be able to define the file mode in the secret. cc @mactsouk @darkofabijan let me know what you think about that.

mactsouk: It is not a bad idea - or maybe keep the file permissions of the original file.
On the other hand, using `chmod` is not a big issue.

PR #48, opened 2018-10-08
Debugging with SSH access

shiroyasha:


debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+To start a debug session for your project, run:
+
+``` bash
+sem debug project <name-of-your-project>
+```
+
+The above command will start a virtual machine connected with your git
+repository and attach you to it via an SSH session.
+
+## Inspect the state of a running job
```
darkofabijan: I would just mention here that they can get a list of running jobs by running `sem get jobs`. I think it would be great usability addition to this section. In that case they really don't need to read any other page to get 98% of all debugging goodness.

darkofabijan:
@shiroyasha Looks great! Once you take care of the links feel free to ship it. I wouldn't change anything at this point. Only if we learn something later from feedback.

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
+Often the best way to troubleshoot failures in your pipelines is to SSH into a
```
mactsouk: I will add a **TOC** once the page is ready. Also, we will need an `Overview` section at the beginning of the page.

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+job and inspect log files, running processes, and directory paths. Semaphore 2.0
+gives you the option to access all running jobs via SSH, to restart your jobs
+in debug mode, or to start on-demand virtual machines to explore the build
```
mactsouk: I would say "the actual build environment".

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+environment.
+
+* [Explore the build environment](#explore-the-build-environment)
+* [Inspect the state of a running job](#inspect-the-state-of-a-running-job)
+* [Restart a job in debug mode](#restart-a-job-in-debug-mode)
+* [Port forward your web server and debug UI issues](#port-forward-your-web-server-and-debug-ui-issues)
+
```
mactsouk: Maybe we can say *Exploring*, *Inspecting*, *Restarting* and *Port forwarding* here. Also, I think that port forwarding is not useful for web servers only but for all kinds of TCP applications.

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+Access to the job's virtual machine is managed by the public SSH keys in the
+`.ssh/authorized_keys` file. To add your public key to the machine add the
+following command to your pipeline definition:
+
+``` bash
+echo '<your-public-key>' >> .ssh/authorized_keys
+```
+
+To manage multiple public keys for SSH access
+[store your public keys in a secret](#link).
```
mactsouk: The link is https://docs.semaphoreci.com/article/61-using-secrets

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Restart a job in debug mode
+
+To find the root cause of a finished failed job, Semaphore allows you to restart
```
mactsouk: "of a finished job that failed" sound better to me.

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+``` bash
+sem debug job <job-id>
+```
+
+This will start a new interactive job based on the specification of the old one,
+export the same environment variables, inject the same secrets, and connect to
+the same git commit.
+
+Commands in the debug mode are not executed automatically, instead they are
+stored in the home directory `/home/semaphore/commands.sh`. This allows you to
```
mactsouk: "stored in a file in the home directory – the name of the file is `/home/semaphore/commands.sh`"

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+
+Semaphore allows you to forward ports to your local machine and explore the UI
+of your application from your browser. If your application is running on port
+`3000`, you can port forward to your local port `3000` with:
+
+``` bash
+sem port-forward <job-id> 3000 3000
+```
+
+The `http://localhost:3000` should now be accessible in your browser.
```
mactsouk: Maybe add a `See also` section - if yes, I will put the necessary links.

debugging-with-ssh-access.md
```diff
@@ -0,0 +1,81 @@
...
+Sometimes SSH access to your build environment is not enough to fully explore
+the problem. For example, Selenium based tests will fail if the html elements
+are not visible on the screen when you are running the tests.
+
+Semaphore allows you to forward ports to your local machine and explore the UI
+of your application from your browser. If your application is running on port
+`3000`, you can port forward to your local port `3000` with:
+
+``` bash
+sem port-forward <job-id> 3000 3000
```
ervinb: It's not entirely clear which port used for what. The first one is probably the local port and the second one is for the remote one.

I think it would help to use different port numbers to differentiate between environments:
```
sem port-forward <job-id> 6000 3000

The `http://localhost:6000` should now be accessible in your local browser.
```

shiroyasha: Good point 🏆

debugging-with-ssh-access.md
shiroyasha: Good point. I opted to not follow this rule in this guide for two reason:

- The first information on the page should give a high level explanation of a concept. Jumping immediately into a TOC delays that information and makes it harder to understand the key points.

Example kubernetes docs: https://kubernetes.io/docs/concepts/configuration/secret/

- The overview section can be very useful for longer documents. I feel that for short guides like this, it is an unnecessary duplication (having both the intro at the begginging of the doc + an overview).

debugging-with-ssh-access.md
shiroyasha: The word `actual` implies that there is an non-actual build environment on Semaphore. Can you give a broader context how the `actual` word makes things clearer?

debugging-with-ssh-access.md
shiroyasha: I agree that port-forwarding can used for much more then simple web server inspection. But, I am not sure if the high level overview is the best place to share this information. 

I want to give concrete examples to issues that arise during development and testing. The best example I have in mind is UI inspection. Everything else can be inspected in the machine with `tcpdump`, `netstat` and similar tools without port-forwarding.

debugging-with-ssh-access.md
shiroyasha: I wanted to avoid making these into a long paragraph that explain details. But I agree with this. 

I'll mention `sem get jobs` as a "quick way to find your job" sentence at the bottom.

debugging-with-ssh-access.md
shiroyasha: Good point. This opened my eyes that I don't even need to say finished, as "failed" already implies that the job is finished and there is no reason to repeat myself.

I'll go with:

> To find the root cause of a failed job

debugging-with-ssh-access.md
shiroyasha: Another good point, I always forget that we are targeting experienced unix users, there is no reason to mention that `/home/semaphore/commands.sh` is a file and that it is storred in the home directory.

I'll go with:

> Commands in the debug mode are not executed automatically, instead they are
stored in `/home/semaphore/commands.sh`.

debugging-with-ssh-access.md
shiroyasha: I wanted to link to a concrete example, but this link will do it for now.

In the future we should add this use case.

PR #51, opened 2018-10-12
Wordsmith Concepts.

markoa:
- Added intro sentence to remind us what Semaphore is.
- Reduced amount of passive voice.
- Some paragraphs split for easier reading.
- Removed use of italics if a concept is not mentioned for the first time.
- Various minor clarifications.
- Added next step section with link.


markoa:
I think the page title should be "Semaphore Concepts" but that's not defined in Markdown.

PR #67, opened 2018-10-17
Updating the cache reference

mimimalizam:


markoa:
I've slightly reworded some parts, merged examples with command reference and changed format to subheader-based, so that explanation text can fit nicely as paragraphs.

PR #72, opened 2018-10-30
Add Python guide

ahawkins:
This follows the same structure as the existing language guides.


language-python.md
```diff
@@ -0,0 +1,106 @@
...
+
+blocks:
+  - name: Warm caches
+    task:
+      env_vars:
+        - name: PIPENV_VENV_IN_PROJECT
+          value: "true"
+      prologue:
+        commands:
+          - sudo pip install pipenv
```
ahawkins: I suggest `pipenv` be installed as part of standard ubuntu image. It's askin to hav having ruby without bundler.

PR #75, opened 2018-10-30
(Deprecated) Using Docker registry for Docker Image reuse

mactsouk:


mactsouk:
@markoa Will do that on Monday - Will do the same for https://github.com/semaphoreci/docs/pull/74.

PR #80, opened 2018-11-01
Add Browser Testing

ahawkins:
A short page on browser testing. I wasn't sure what else to add to this page because everything I tested just worked out of the box! 🎉 on that team Semaphore. I figured it would be enough to echo the fact that the standard selenium setup (with heads and headless modes) should work out of the box. Any other suggestions?


markoa:
Thanks @ahawkins! I've distributed this content in a way that we actually don't create a new page, in #104. It's hard to know what to say when things just work :) so we'll iterate later based on feedback.

PR #114, opened 2018-11-05
Updating cache toolbox reference

mimimalizam:


toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -215,9 +215,14 @@ Examples:
     cache store gems-$SEMAPHORE_GIT_BRANCH-revision-$(checksum Gemfile.lock) vendor/bundle
 
 Archives file or directory specified by path and associates it with key.
```
mactsouk: I would rephrase it as:

The `cache store` command archives files and/or directories specified by `path` and associates them with the given `key` value.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -215,9 +215,14 @@ Examples:
...
-Any further changes of path after the store operation completes will not
+As `cache store` uses `tar` leading `/` is automatically removed from the path.
```
mactsouk: I would rephrase it as:

As `cache store` uses the `tar` utility, it automatically removes any leading `/` (if they exist) from the given `path` value.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -215,9 +215,14 @@ Examples:
...
+Also, any further changes of path after the store operation completes will not
 be automatically propagated to cache. The command always passes (exits with 0).
 
+Allocated disk space for cache is currently limited to 9.6G.
```
mactsouk: I would rephrase it as:

The allocated disk space...

Question: is that cache space allocated _per user_ or _per organization_?

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -215,9 +215,14 @@ Examples:
...
 Archives file or directory specified by path and associates it with key.
-Any further changes of path after the store operation completes will not
+As `cache store` uses `tar` leading `/` is automatically removed from the path.
+Also, any further changes of path after the store operation completes will not
 be automatically propagated to cache. The command always passes (exits with 0).
 
+Allocated disk space for cache is currently limited to 9.6G. 
+Also, in case of insufficient disk space, command `cache store` allocates space
+for new archive by deleting the oldest keys. 
+
```
mactsouk: I would rephrase it as:

Also, in case of insufficient disk space, the `cache store` command frees disk space by deleting the
oldest keys.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -289,6 +297,8 @@ which are automatically set in every job environment:
     SSH key that will be used for connecting to the cache server
   (`/home/semaphore/.ssh/semaphore_cache_key`).
 
+Additionally, `cache` uses `tar` to archive a specified directory or file.
```
mactsouk: I would rephrase it as (in order to be more generic):

Additionally, `cache` uses `tar` to archive the specified directories and files.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -229,6 +234,9 @@ Examples:
 Restores an archive with given key.
```
mactsouk: Maybe: "Restores the archive with the given key".

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
```diff
@@ -229,6 +234,9 @@ Examples:
...
 In case of a cache hit, archive is retrieved and available at its original
 path in the job environment.
+Cached directory (or file) is restored on the current path,
```
mactsouk: I would say "The cache directory..."

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mimimalizam: Cache store is created per project basis and also disk size is allocated per project.

mactsouk: I did not know that :) I think you should mention that information on the page.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mimimalizam: Cache store command can be used in the following examples:

- To cache directory we specify directory path. For example, if we want to cache `vendor/bundle` we'd use a command similar to the following:
`cache store gems vendor/bundle`
- On the other hand, we can also cache only a single file:
`cache store gemfile-list Gemfile.lock`

I'm under the impression that stating `files` in this context will create confusion - that one execution of `cache store` can store more files at once.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mimimalizam: Sorry for the confusion, this part is stated in the description paragraph.

```
Cache is created on a per project basis and available in every pipeline job. All cache keys are scoped per project.
```

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mimimalizam: 🤔 Tnx, but `cache directory` sounds a bit vogue to me. 

With `cached directory` I wanted to stress out that it is related to a specific archive stored by a specific key.

I'm worried that `cache directory` can give the impression that it is the whole cache directory for project, while in this case it is just one archive.

What do you think, would `An archive is restored on the current path` make more sense? cc @mactsouk @darkofabijan

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mactsouk: My fault – I always edit for clarity but this time it was not necessary :)

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mactsouk: Maybe use `Each archive` instead of `An archive`.

mimimalizam: 🤔 I'm still in doubt, thing is that `cache restore` restores only one archive when it is called.

mactsouk: So, what about saying:
`Each archive, either directory or single file, is restored in the current path`.
Does this make more sense?

mimimalizam:
Tnx for the extensive feedback! I've partially updated the changes proposed by this PR. For the time being, since command `cache store` (`cache restore`) stores (restores) only one archive when it is called - in order to avoid confusion, I decided to use `directory` and `file` instead of `directories` and `files` in examples. But I'll leave the final decision to @mactsouk and @darkofabijan.

toolbox-reference_5b7bdc1d0428631d7a8a2fcc.md
mimimalizam: Description of cache states that an archive can be a single file or directory. So, I decided to go with `each archive in this example`.

```
cache tool uses key-path pairs for managing cached archives. An archive can be a single file or a directory.
```

markoa:
@mimimalizam @mactsouk what's the status of this PR / next steps?

markoa:
P.S. I can take over if some more work is needed.

mimimalizam:
I think we should get another pair of eyes on the `cache` reference. If you have the time, I'd appreciate if you could give it a go. 🙇 cc @markoa

markoa:
Made some light edits, thanks!

PR #125, opened 2018-11-16
Provide a recipe for generating database.yml in Ruby guide

### LINKED ISSUE #112: Provide a recipe for generating database.yml in Ruby guide
Semaphore Classic autogenerates this file for every Ruby on Rails project. Let's make transition easier by having an example they can copy and paste.

ahawkins:
Close #112

### DIFF
--- a/language-ruby_5bb9e082042863158cc7236d.md
+++ b/language-ruby_5bb9e082042863158cc7236d.md
@@ -95,11 +95,17 @@ Semaphore doesn't set language specific environment variables like
 
 ## Rails
 
-Configuring Rails applications require a few extra steps. You'll need
-to set `RAILS_ENV` like discussed earlier and generate a
-`config/database.yml`. The simplest way to start a database with
-`sem-service`, set `DATABASE_URL`, and write a new
-`config/database.yml`. Here's an example block:
+You'll need to set `RAILS_ENV` and configure a databse. Rails 4.1+
+supports using the `DATABASE_URL` to [configure a
+database][rails-database-configuration]. Older versions of Rails need
+to generate a `database/config.yml` manually.
+
+### 4.1+ - Using DATABASE_URL
+
+Rails will read database configuration from `DATABASE_URL` if set and
+`config/database.yml` is empty. So, start a database with
+[sem-service][sem-service], set `DATABASE_URL`, and remove the
+existing `config/database.yml`. Here's an example block:
 
 <pre><code class="language-yaml">blocks:
   - name: Tests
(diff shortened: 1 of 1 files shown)


language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -92,6 +93,34 @@ Semaphore doesn't set language specific environment variables like
             - bundle exec rake test
 </code></pre>
 
+## Rails
+
+Configuring Rails applications require a few extra steps. You'll need
```
markoa: This sentence can be removed without effect, so let's remove it. We can assume that many readers will jump straight to this section. And always prefer opening in a positive tone, "xyz requires extra steps" sounds like ok this isn't supported but here's a workaround.

language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -92,6 +93,34 @@ Semaphore doesn't set language specific environment variables like
...
+to set `RAILS_ENV` like discussed earlier and generate a
+`config/database.yml`. The simplest way to start a database with
+`sem-service`, set `DATABASE_URL`, and write a new
```
markoa: link to sem-service reference

language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -92,6 +93,34 @@ Semaphore doesn't set language specific environment variables like
             - bundle exec rake test
 </code></pre>
 
+## Rails
+
+Configuring Rails applications require a few extra steps. You'll need
+to set `RAILS_ENV` like discussed earlier and generate a
+`config/database.yml`. The simplest way to start a database with
```
markoa: I think that the example can be made more clear:
- `DATABASE_ENV` is a special env var that Rails understands, but the example seems to use it as random temporary storage variable
- Rails 4.1+ can read db config from from `DATABASE_ENV` only
- However older versions need `config/database.yml`

I suggest providing two examples, one for 4.1+ with `DATABASE_ENV`, and another with `config/database.yml` which is stored in git and cp-ed in prologue. I suggest we show users to store such Semaphore-specific files in `.semaphore` folder, along with pipeline definition(s).

ahawkins:
@markoa addressed your feedback

language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -92,6 +93,81 @@ Semaphore doesn't set language specific environment variables like
             - bundle exec rake test
 </code></pre>
 
+## Rails
+
+You'll need to set `RAILS_ENV` and configure a databse. Rails 4.1+
+supports using the `DATABASE_URL` to [configure a
+database][rails-database-configuration]. Older versions of Rails need
```
markoa: This link is undefined...

language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -92,6 +93,81 @@ Semaphore doesn't set language specific environment variables like
             - bundle exec rake test
 </code></pre>
 
+## Rails
+
+You'll need to set `RAILS_ENV` and configure a databse. Rails 4.1+
```
markoa: typo

language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -92,6 +93,81 @@ Semaphore doesn't set language specific environment variables like
...
+supports using the `DATABASE_URL` to [configure a
+database][rails-database-configuration]. Older versions of Rails need
+to generate a `database/config.yml` manually.
+
+### 4.1+ - Using DATABASE_URL
+
+Rails will read database configuration from `DATABASE_URL` if set and
+`config/database.yml` is empty. So, start a database with
+[sem-service][sem-service], set `DATABASE_URL`, and remove the
+existing `config/database.yml`. Here's an example block:
```
markoa: Let's just add a note that this is optional and applies if you have a `config/database.yml` in Git.

ahawkins: It's more accurate to say if your `config/database.yml` is not blank. My guess is the majority of cases have `config/database.yml` in SCM (since it's the rails default) and you need the file to configure dev and test bases. It seems that using `DATABASE_URL` is primarily for deployments (or CI in our case).

markoa: Agreed.

PR #128, opened 2018-11-16
Add pip caching example

### LINKED ISSUE #88: Provide info on using pip
Current Python docs show use of pipenv, which is fairly new and not preinstalled.

Pip is preinstalled and common so let's have show how to use pip in dependency caching section first, before pipenv.

ahawkins:
Close #88 

Perhaps a section on working with virtualenv's is in order?

### DIFF
--- a/language-python_5bdc36652c7d3a01757aba24.md
+++ b/language-python_5bdc36652c7d3a01757aba24.md
@@ -47,7 +47,7 @@ example:
           - checkout
           - mkdir .pip_cache
           - cache restore v1-pip-$(checksum requirements.txt)
-          # User --user to avoid permission conflicts
+          # Use --user to avoid permission conflicts
           - pip install --user --cache-dir .pip_cache -r requirements.txt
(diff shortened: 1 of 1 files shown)


language-python_5bdc36652c7d3a01757aba24.md
```diff
@@ -30,6 +30,34 @@ Python 2.7 is the default version. This can be switched to 3.7 with
...
+  - name: Tests
+    task:
+      prologue:
+        commands:
+          # Add --local pip bin to $PATH
+          - export PATH=$HOME/.local/bin:$PATH
+          - checkout
+          - mkdir .pip_cache
+          - cache restore v1-pip-$(checksum requirements.txt)
+          # User --user to avoid permission conflicts
```
markoa: Is this meant to be "Use --user", as in install as user?

markoa:
@ahawkins pls see the tiny question and then we can merge.

re. virtualenv, I'm not a Pythonista but it sounds like some pointers would be helpful. Feel free to submit a PR.

PR #129, opened 2018-11-17
Caching Big Repositories recipe

### LINKED ISSUE #123: Document recipe for caching big repositories

mactsouk:
Using Semaphore 2.0 cache server instead of `checkout`.

Closes #123.

### DIFF
--- a/caching-big-repositories_5bf03fb92c7d3a31944e139d.md
+++ b/caching-big-repositories_5bf03fb92c7d3a31944e139d.md
@@ -34,14 +34,19 @@ remain intact during the lifetime of the Semaphore 2.0 project.
 
 ## An example project
 
-The example project of this section will use a single large file.
+The example project of this section will use multiple large files. Each one of
+the 10 files is around 100MB in size.
+
+Each file was generated using the following command:
+
+    dd if=/dev/random of=filename bs=1024 count=100000
 
 The `semaphore.yml` file for the example project will be the following:
 
 
 The Semaphore 2.0 environment variables that will be used are the following:
 
-* `SEMAPHORE_GIT_REPO_NAME`:
+* `SEMAPHORE_PROJECT_NAME`:
 * `SEMAPHORE_GIT_SHA`:
 
 ## Evaluating the results
--- a/caching-big-repositories_5bf03fb92c7d3a31944e139d.md
+++ b/caching-big-repositories_5bf03fb92c7d3a31944e139d.md
@@ -1,35 +1,24 @@
 
 * [Overview](#overview)
-* [The scenario](#the-scenario)
 * [The logic behind the recipe](#the-logic-behind-the-recipe)
 * [An example project](#an-example-project)
 * [See Also](#see-also)
 
 ## Overview
 
-The purpose of this document is to provide a recipe for caching big
-repositories in order to save time.
-
-## The scenario
-
 Imagine that you have one or multiple large files on a GitHub repository and
-that you want to make the downloading of them as fast as possible.
-
-## The logic behind the recipe
-
-Using the Cache server provided by Semaphore 2.0 for getting the files of your
-GitHub repository is much faster than using the GitHub server machines for the
-same task because the Semaphore Cache server is on the same network as the
-Virtual Machines used for executing the jobs of a Semaphore pipeline.
-
-What the example project that follows will do is getting all the files of the
-GitHub repository in the first job using `checkout` and then storing them into
-the Semaphore 2.0 Cache server using `cache store`. The remaining jobs of the
-project will use these files from the Cache server using `cache restore`
-without calling `checkout` again.
+you want to make the downloading of them as fast as possible.
 
 Note that for this recipe to work, the files of the GitHub repository should
 remain intact through the lifetime of the Semaphore project.
+
+## The logic behind the recipe
+
+The example project that follows is getting all the files of the GitHub
+repository in the first job using `checkout` and then is storing them into the
+Semaphore 2.0 Cache server using `cache store`. The remaining jobs of the
+project will use these files from the Cache server using `cache restore`
+without calling `checkout` again.
 
 ## An example project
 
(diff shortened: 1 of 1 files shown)


caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,167 @@
...
+* [The logic behind the recipe](#the-logic-behind-the-recipe)
+* [An example project](#an-example-project)
+* [See Also](#see-also)
+
+## Overview
+
+The purpose of this document is to provide a recipe for caching big
+repositories in order to save time.
+
+## The scenario
```
markoa: This section and the overview should be one section without a subheader. Just jump straight to the purpose of the guide.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,167 @@
...
+
+Imagine that you have one or multiple large files on a GitHub repository and
+that you want to make the downloading of them as fast as possible.
+
+## The logic behind the recipe
+
+Using the Cache server provided by Semaphore 2.0 for getting the files of your
```
markoa: What you say in these 3 paragraphs can be said in one short paragraph.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,167 @@
...
+the Semaphore 2.0 Cache server using `cache store`. The remaining jobs of the
+project will use these files from the Cache server using `cache restore`
+without calling `checkout` again.
+
+Note that for this recipe to work, the files of the GitHub repository should
+remain intact through the lifetime of the Semaphore project.
+
+## An example project
+
+The example project of this section will use 10 large files. Each one of these
```
markoa: We don't need to provide any details about example project. Proceed straight to a configuration that people can copy and paste.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,167 @@
...
+	  - name: Task 5 - Single Job
+	    task:
+	      jobs:
+	      - name: Use Cache - Job 5
+	        commands:
+	          - cache restore "$SEMAPHORE_PROJECT_NAME-$SEMAPHORE_GIT_SHA"
+	          - ls -al "$SEMAPHORE_PROJECT_NAME"
+	          - du -k -s
+
+## Evaluating the results
```
markoa: We don't need a section which compares performance.

markoa:
@predrag-rakic Earlier version of page (still available within this PR) had results. I asked Mihalis to remove them. There was some improvement, but it was also the week of network hiccups, so it might not be the final number. Note that it's also about higher resiliency and avoiding transcontinental roundtrips.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,105 @@
+
+* [Overview](#overview)
+* [The logic behind the recipe](#the-logic-behind-the-recipe)
+* [An example project](#an-example-project)
+* [See Also](#see-also)
+
+## Overview
+
+Imagine that you have one or multiple large files on a GitHub repository and
```
markoa: This is technically not fully correct, eg more than half of a Git repo's size can be in history of past files, some of which were big. Anyway it is beyond scope of this guide to discuss why a Git repo can be large. It needs to start from the fact that somebody just has a large Git repo.

So we can streamline the intro by saying something like:

_This guide shows you how to cache a large Git repository and avoid cloning it from scratch in every block._

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,105 @@
...
+you want to make the downloading of them as fast as possible.
+
+Note that for this recipe to work, the files of the GitHub repository should
+remain intact through the lifetime of the Semaphore project.
```
markoa: I'm surprised by this. Are you sure?

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,105 @@
...
+
+## The logic behind the recipe
+
+The example project that follows is getting all the files of the GitHub
+repository in the first job using `checkout` and then is storing them into the
+Semaphore 2.0 Cache server using `cache store`. The remaining jobs of the
```
markoa: I think that this recipe should do more than this.

- Keep the repo in cache throughout workflows so that even the first block doesn't need to do `git clone`;
- Pull the latest changes in requested branch only.
- This all assumes cache hit, but there should be a fallback in the first block in case of cache miss.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
mactsouk: I should have said `lifetime of pipeline` instead.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
mactsouk: Done.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
mactsouk: I would suggest that we keep that recipe _simple_ and maybe create a new recipe with a more complex scenario.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,107 @@
...
+    
+	  - name: Task 2 - Single Job
+	    task:
+	      jobs:
+	      - name: Use Cache - Job 1
+	        commands:
+	          - cache restore "$SEMAPHORE_PROJECT_NAME-$SEMAPHORE_GIT_SHA"
+	          - ls -al "$SEMAPHORE_PROJECT_NAME"/FILES
+    
+	  - name: Task 3 - Single Job
```
markoa: would reader lose some understanding if we removed blocks 3, 4 and 5?

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
predrag-rakic: I honestly can't understand this.
What is the message this paragraph is conveying?
What does it mean "repository should remain intact"?
What is the scenario that users need to protect themselves against and how?

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
predrag-rakic: And add at least one more job to Task 2.
Otherwise the example is meaningless.

As a side-note: entities named "Task #" are Blocks and should be named that to avoid confusion.

predrag-rakic:
I would suggest a paragraph to set the expectations for this optimization:
- it will probably improve stability
- it will probably reduce price (total number of seconds all jobs in a pipeline combined spend)
- if will probably not decrease total pipeline execution time.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
mactsouk: Renamed `Task #` to `Block item` and added another Block item at the end.

caching-big-repositories_5bf03fb92c7d3a31944e139d.md
```diff
@@ -0,0 +1,107 @@
...
+    
+	blocks:
+	  - name: Reuse GitHub repository
+	    task:
+	      jobs:
+	      - name: Cache GitHub repository
+	        commands:
+	          - checkout
+	          - cd ..
+	          - cache store "$SEMAPHORE_PROJECT_NAME-$SEMAPHORE_GIT_SHA" $SEMAPHORE_PROJECT_NAME
```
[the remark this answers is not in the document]
mactsouk: We can add `cache delete` as an `epilogue` command to the last `Block item`.

mactsouk:
@predrag-rakic Adding a paragraph to set the expectations is a great idea. Done.

markoa:
OK so we spent a lot of time on this topic and in the end decided to make git repo caching a product feature. Work is already under way. So this guide will be obsoleted in a matter of days -> closing.

Issue #130, opened 2018-11-18
Ruby caching example missing info for restoring cache


markoa:
A user reported:

> In your Ruby docs you provide a demonstration on how to cache your gems in vendor/bundle and then use the BUNDLE_PATH environment in the next task to tell bundler where to find the dependencies. However I couldn’t get this to work at all like the docs described - I had to run `bundle check --path vendor/bundle` after restoring the cache to get bundler to recognise the dependency location. Is this an issue with the docs or something I’m doing?

PR #134, opened 2018-11-19
Correct caching example

ahawkins:
The .bundle directory must be cached and restored as well since that saves settings after bundle install --deployment.

Close: #130.

Refer to https://github.com/ahawkins/semaphore2-example-rails for a working implementation.

### DIFF
--- a/language-ruby_5bb9e082042863158cc7236d.md
+++ b/language-ruby_5bb9e082042863158cc7236d.md
@@ -32,10 +32,10 @@ Here's an example:
 
 ## Dependency caching
 
-You can use Semaphores `cache` command to store and load a gem bundle.
-This requires setting the `BUNDLE_PATH` environment variable. In the following
-configuration example, we install dependencies and warm the cache in the first
-block, then use the cache in subsequent blocks.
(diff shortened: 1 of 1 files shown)


language-ruby_5bb9e082042863158cc7236d.md
```diff
@@ -52,19 +52,18 @@ blocks:
...
   - name: Tests
     task:
       prologue:
         commands:
           - checkout
-          - cache restore bundle-$(checksum Gemfile.lock)
-      env_vars:
-        - name: BUNDLE_PATH
-          value: vendor/bundle
+          - cache store bundle-gems-$(checksum Gemfile.lock)
```
markoa: store or restore?

ahawkins: Fixed.

darkofabijan:
@markoa @ahawkins would switching command from:

`bundle install --deployment` to `bundle install --deployment --path vendor/bundle` work? It seems much easier than storing and restoring two directories to/from cache.

markoa:
That would be better.

PR #143, opened 2018-11-24
Services are not shared

mactsouk:


markoa:
Made some minor edits to adapt for the wider context of the pages.

PR #196, opened 2018-12-04
Docker image recipe (indefinitely in cache)

### LINKED ISSUE #172: How to cache parent or base Docker images
Scenario: my project is using some Docker images as a starting point. For example my service runs on golang:alpine [parent image](https://docs.docker.com/glossary/?term=parent%20image) and also connects to a [mongodb](https://hub.docker.com/_/mongo/) base image via docker compose. I don’t want to pull my parent or dependency images from scratch in every workflow. These images should stay in cache "indefinitely".

mactsouk:
Closes #172


markoa:
After internal discussion, we won't recommend this as primary approach in docs.

PR #205, opened 2018-12-06
'sem create -f' and 'on-demand job creation'

mactsouk:


shiroyasha:
It would be useful to point out that this is a plumbing level command in sem CLI, mainly used in other high level commands like `sem debug project <id>` and `sem debug job <id>`. You can, of course, use it directly if you need some very specific needs, but most people would use it through `sem debug`.

This would answer the question: "What is the point of this command?".

---

Edit, Just to clarify:

`sem debug project` is mostly equivalent to the following pseudocommands:

```
id=$(sem create job "debug session" --command "sleep 3600" --project <name> | awk -F "'" '{ print $2 }'

while ! (sem get job $id | grep "running"); do
  echo "Waiting for the job to start"
done 

sem attach $id
```

markoa:
@shiroyasha Well I don't see why a user would want to run this command ever. What do you think about removing it from user documentation completely? Maybe it's not "clean" but I'd optimize for clarity and ease of understanding.

shiroyasha:
@markoa I agree, lets remove it.

shiroyasha:
Just to expand with some further information.

In the future, this command can be a useful addition to a developers local developing environment.

Here is one of my current use cases for example: I spin up a Kubernetes cluster with more memory than my local machine (e1-standard-8) and test out various scaling policies in Istio.

It is fast, dirty, and when I finish I do `sem stop job <id>`.

PR #240, opened 2018-12-17
Add Django section

### LINKED ISSUE #185: CI guide for Django
Use the [Rails guide](https://docs.semaphoreci.com/article/99-rails-continuous-integration) as a blueprint.

ahawkins:
I added a short section on configuring the database for a Django project. I tested this out on a sample project and it worked well enough. Are any other topics required?

Close #185

### DIFF
--- a/use-case-django-continuous-integration.md
+++ b/use-case-django-continuous-integration.md
@@ -17,7 +17,7 @@ We want fast feedback through the test pipeline, so we'll fail fast if
 the linting fails. This prevents running time consuming tests for
 semantically incorrect code.
 
-Consider a smaple Django `blog` project with two applications: `posts`
+Consider a sample Django `blog` project with two applications: `posts`
 and `comments`. For these reasons our pipeline is composed of three
 blocks, one for setup and two for tests:
 
@@ -37,9 +37,11 @@ blocks:
           commands:
             - sem-version python 3.7
             - checkout
-            - cache restore requirements-$(checksum requirements.txt)
+            # Partial cache key matching ensures that new branches reuse gems
+            # from the last build of master.
+            - cache restore requirements-$SEMAPHORE_GIT_BRANCH-$(checksum requirements.txt),requirements-$SEMAPHORE_GIT_BRANCH-,requirements-master-
             - pip download --cache-dir .pip_cache -r requirements.txt
-            - cache store requirements-$(checksum requirements.txt) .pip_cache
+            - cache store requirements-$SEMAPHORE_GIT_BRANCH-$(checksum requirements.txt) .pip_cache
 
   - name: Lint
     task:
@@ -47,7 +49,7 @@ blocks:
         commands:
           - sem-version python 3.7
           - checkout
-          - cache restore requirements-$(checksum requirements.txt)
+          - cache restore requirements-$SEMAPHORE_GIT_BRANCH-$(checksum requirements.txt)
           - pip install -r requirements.txt --cache-dir .pip_cache
       jobs:
         - name: Project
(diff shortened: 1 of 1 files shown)


markoa:
@ahawkins The idea behind all "CI guide for X" is to provide a complete production-grade configuration that people can easily copy and paste (aka guide for the impatient ;). That's why we have a unique page on Rails, even though you could piece all that together from reading the Ruby page + reference docs.

So please create a new page laying out a pipeline similar to the Rails one (eg `django.md`, I'll rename it later for publishing in CMS), even though there'll be a bit of overlap with the Python page.

use-case-django-continuous-integration.md
```diff
@@ -0,0 +1,148 @@
...
+This example configures coverage.py and and pylint. Be sure to exclude
+the relevant files from coverage by creating a `.coveragerc`:
+
+<pre><code class="language-conf"># .coveragerc
+[run]
+source = .
+omit =
+  venv/*
+  manage.py
+  */wsgi.py
```
ahawkins: Do we need to escape these for proper parsing?

ahawkins:
@markoa I think this is what you had in mind.

use-case-django-continuous-integration.md
```diff
@@ -0,0 +1,148 @@
...
+
+blocks:
+  - name: Setup
+    task:
+      jobs:
+        - name: Cache pip downloads
+          commands:
+            - sem-version python 3.7
+            - checkout
+            - cache restore requirements-$(checksum requirements.txt)
```
markoa: Would pip benefit from restoring a previous version of deps bundle? If so let's add fallback keys (see Rails guide for example) — cache restore does partial matching.

use-case-django-continuous-integration.md
```diff
@@ -0,0 +1,148 @@
...
+          value: blog.ci
+      prologue:
+        commands:
+          - sem-service start postgres
+          - sem-version python 3.7
+          - checkout
+          - cache restore requirements-$(checksum requirements.txt)
+          - pip install -r requirements.txt --cache-dir .pip_cache
+      jobs:
+        # Create a job for each application in the project
```
markoa: Let's mention that they're "parallel jobs".

use-case-django-continuous-integration.md
markoa: Shouldn't be necessary. I'll check after publishing.

use-case-django-continuous-integration.md
```diff
@@ -0,0 +1,148 @@
...
+Finally, you'll want to install `pylint-django` and add to the default
+plugin list. Here's a snippet for `.pylintrc`:
+
+<pre><code class="language-conf"># .pylintrc
+load-plugins=pylint_django
+</code></pre>
+
+## Next steps
+
+Congratulations! You have set up your first Rails 5 continuous integration
```
markoa: Rails 5

use-case-django-continuous-integration.md
```diff
@@ -0,0 +1,148 @@
...
+
+Mature applications usually have many tests and optimizing their runtime saves
+a lot of valuable time for development. So for demonstration we'll run
+tests for each application in the project in parallel.
+
+We want fast feedback through the test pipeline, so we'll fail fast if
+the linting fails. This prevents running time consuming tests for
+semantically incorrect code.
+
+Consider a smaple Django `blog` project with two applications: `posts`
```
markoa: typo

ahawkins:
@markoa back to you.

markoa:
@ahawkins  Thank you!

PR #250, opened 2018-12-21
Erlang Language Guide

mactsouk:


language-erlang_5c1bb3cc04286304a71d943e.md
```diff
@@ -0,0 +1,100 @@
...
+## Supported Erlang versions
+
+The list of supported Erlang versions is the following:
+
+* 20.3
+* 21.0
+
+## Changing Erlang version
+
+The `kerl` Erlang utility can help you change the Erlang version.
```
bogyo210: Just to be clear here, i suggest : 
"To change erlang version we provide 'sem-version erlang [erlang_version]' utility which is based on 'kerl'  "

PR #253, opened 2018-12-21
Uses APT mirror

### LINKED ISSUE #252: Document that we are using APT mirror and that caching packages has little effect
Probably worth mentioning on Ubuntu 1804 page

mactsouk:
Closes #252

### DIFF
--- a/ubuntu-1804-image_5b461f730428630abc0bf442.md
+++ b/ubuntu-1804-image_5b461f730428630abc0bf442.md
@@ -15,13 +15,15 @@
 * [Ruby](#ruby)
 * [See also](#see-also)
 
-
 ## Overview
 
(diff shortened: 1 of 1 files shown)


ubuntu-1804-image_5b461f730428630abc0bf442.md
```diff
@@ -15,14 +15,16 @@
...
-
 ## Overview
 
 The `ubuntu1804` image is based on [Ubuntu 18.04 LTS](https://wiki.ubuntu.com/BionicBeaver/ReleaseNotes).
 The user in the environment is named `semaphore`, and has full `sudo` access.
 
 Image name: `ubuntu1804`
 
+Currently the `ubuntu1804` Virtual Machine uses an *APT mirror*, which means that
+caching packages will have little effect.
```
darkofabijan: The `ubuntu1804` Virtual Machine uses an *APT mirror* which is in the same data center as our build cluster, which means that caching packages will have little effect.

PR #258, opened 2018-12-24
Use sem-version instead of rbenv

### LINKED ISSUE #257: Clarification on rbenv
https://docs.semaphoreci.com/article/32-ubuntu-1804-image#ruby
There says that rbenv is used to manage local ruby versions. This is confusing because only 3 packages are available to rbenv. The sem-version tool should be used always, because this has access to precompiled versions also.

mactsouk:
Closes #257

### DIFF
--- a/ubuntu-1804-image_5b461f730428630abc0bf442.md
+++ b/ubuntu-1804-image_5b461f730428630abc0bf442.md
@@ -174,7 +174,7 @@ Supporting libraries:
 
 ### Ruby
 
-Ruby versions are installed and managed by [rbenv](https://github.com/rbenv/rbenv).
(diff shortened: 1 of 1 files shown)


ubuntu-1804-image_5b461f730428630abc0bf442.md
```diff
@@ -174,7 +174,7 @@ Supporting libraries:
 
 ### Ruby
 
-Ruby versions are installed and managed by [rbenv](https://github.com/rbenv/rbenv).
+Ruby versions are installed and managed by [sem-version](https://docs.semaphoreci.com/article/54-toolbox-reference#sem-version).
```
bogyo210: Ruby versions are managed by [sem-version]. No need to mention installed

PR #296, opened 2019-02-11
Introduce Apple hardware

shiroyasha:
This PR is a preparation for the beta release of the MacOS build platform.

- What do you think of the separation between Linux & Apple machine types? This is technically not correct, there is no machine type called "linux"? An alternative would be to call them "Standard" & "Apple".
- Is it better like this, or should we keep them in the same table?

---

TODO: specify Apple hardware specs in the documentation (replace ? with real values)


shiroyasha:
Note: Removed the precise hardware specs for mac machines for the time being. They'll be added in later.

PR #360, opened 2019-04-12
Tour with multiple types of environments

shiroyasha:
To keep the Build Environment section easier to understand, and to make sure that we treat all environments with the same respect, I've done some changes in the tour.

- First, the Build Environment page now mainly talks about Linux vs. MacOS vs. Compose.
- I've introduced a new page for Databases and Services

This makes the overall Guide longer but reduces the per-page complexity of the articles.


tour4-build-env_5bb917152c7d3a04dd5b5d66.md
```diff
@@ -1,7 +1,61 @@
-Semaphore runs each job in a clean and isolated environment.
-The default [Ubuntu environment][ubuntu] includes common tools needed by most
-projects. With unrestricted `sudo` access and fast network, you can
-install additional software and customize the environment however you need.
+Semaphore runs each job in a clean and isolated environment. Three types of
+build environments are available:
+
+- The default [Linux Ubuntu environment][ubuntu] that includes common tools
```
markoa: Remove the word "default"

tour4-build-env_5bb917152c7d3a04dd5b5d66.md
```diff
@@ -1,7 +1,61 @@
...
+  needed by projects. It offers full `sudo` access to the virtual machine.
+
+- The [MacOS environment][macos] that includes common tools needed for iOS
+  development. It offers full `sudo` access to the virtual machine.
+
+- The containerized [Docker build environment][docker] that allows you to run
```
markoa: I think we can just call it "Docker environment"?

markoa:
Do you plan more updates to tour or this is all? (I think we discussed updating a few more sections)

shiroyasha:
My next step is to allow/document on how to pull private docker images. Nothing else, in particular, is on my list when it comes to Docker support in the docs. Did I miss something? 🤔

PR #361, opened 2019-04-15
Updated django-continuous-integration use case

TomFern:
To match Django demo project: 

https://github.com/semaphoreci-demos/semaphore-demo-python-django


markoa:
Excellent! 

Just one detail, we can't host images in this repository, so please remove the png and make the image link point to the image in the demo project (see other guides for an example).

TomFern:
Thanks! Removed the png from this PR and changed linked image to the demo project repo.

PR #366, opened 2019-04-17
Create secret with Env Vars the same way you would inject an environment variable into Docker

shiroyasha:
Example of newly available flags in Sem CLI `v0.10.0`:

```
sem create secret aws-secrets -e AWS_KEY=123
```

This is the same way you would inject an env var into Docker, so it should be familiar to a wide range of customers.

I've updated:

- the Secrets Guide. Now we prefer using flags directly over creating/editing files.
- the Sem reference document

---

This PR and the new feature are a preparetion for pulling private Docker image. There the docs will say:

1 ) Please create a secret to hold your docker keys

``` bash
sem create secret dockerhub-pull-secrets \
  -e DOCKER_CREDENTIAL_TYPE=DockerHub \
  -e DOCKERHU_PASSWORD=a \
  -e DOCKERHUB_USERNAME=b
```

2 ) Use this secret to pull images:

```
agent:
  machine:
    type: e1-standard-2

  containers:
     - name: main
        image: renderedtext/hello

   image_pull_secrets:
     - name: dockerhub-pull-secrets
```

### DIFF
--- a/docker-dockerhub_5bb9dcf5042863158cc72368.md
+++ b/docker-dockerhub_5bb9dcf5042863158cc72368.md
@@ -35,7 +35,7 @@ blocks:
     task:
       # Pull in environment variables from the "docker" secret
       secrets:
-        - name: docker
+        - name: dockerhub-secrets
       prologue:
         commands:
           # Authenticate to the registry for all jobs in the block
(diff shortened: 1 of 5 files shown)


shiroyasha:
Everything updated. Thanks for the review. Merging.

PR #369, opened 2019-04-17
Move sem-service and sem-version to ubuntu1804

shiroyasha:
The motivation for this:

The sem-service and sem-version utility tools only work on Ubuntu 18.04 virtual machines. In this regard, they are totally different from other tools in our toolbox like `retry`, `checkout`, `cache`, etc... that work cross-platform and also in any Docker image.

In the future, we can investigate the possibility of moving the source code of these utilities outside of toolbox, and ship it baked in into our VM images.

---

TODO: some other parts of the docs might need to be adjusted as well. I'll investigate before merging.

---

Pinging also @darkofabijan @CommanderK5 @bogyo210 to verify this change in our toolbox approach. I'm looking forward to your opinion on this topic.


markoa:
👍 for making it clear that sem-service and sem-version are Ubuntu-image only.

In example projects source code, documentation and blog posts we often need to link to sem-service when we mention it for the first time. So right now there many links pointing to `toolbox#sem-service` that need to be adjusted.

Going forward I think it'll be more practical to have them have their own pages, instead changing to `somethingelse##selecting-lanaguge-version-with-sem-version` and adding a lot of new sub-content to the Ubuntup age. sem-service and sem-version would be listed and briefly explained on the Ubuntu page.

shiroyasha:
I see your point about a separate page for sem-service and sem-version.
One thing that I don't yet know is how to call that page.

One idea is: "Ubuntu 18.04 toolbox" but it doesn't resonate with me, or is it something that our customers would search for.

Alternatively, we can have two new pages called:

1. Ubuntu 18.04: Managing Language Versions
2. Ubuntu 18.04: Managing Databases and Background services

The second page could contain all the content currently present on:
https://docs.semaphoreci.com/article/118-configuring-database-access

---

At this point, neither of my proposed options resonate well with me. @markoa do you maybe have an alternative suggestion?

markoa:
The pages would be:

"sem-version: Managing language version on Linux"
"sem-service: Managing databases and services on Linux"

shiroyasha:
@markoa I've introduced the new pages into the docs.

Is there a way to get the links to the pages before I merge to master? So far I've used placeholder links to connect the documents.

shiroyasha:
Thanks for the explanation. Also, a good idea to put it in the readme 👏

PR #379, opened 2019-04-24
Add separate page for Conditions DSL

DamjanBecirovic:
Here is a separate page for Conditions. Someone will need to create it n HelpScout once it is merged.

Feedback is appreciated, especially on `Introduction` part.

Also, I'm not sure about number of examples and whether they should be full yaml files, or just snippets like here, or if some dots should be added to indicate missing parts.


DamjanBecirovic:
@markoa Thanks a lot for review, especially for Intro part, I struggled a bit with tone there and the proposition is way better.

I changed the intro part, fixed formatting everywhere except in tables because newline breaks them, and adjusted examples titles. 

Also `result` and `result_reason` were removed from formal spec section because they are only used for auto-promotion which does not support when conditions yet.

I will readd them once that is complete.

conditions-dsl.md
```diff
@@ -0,0 +1,147 @@
...
+|   !~      | True if keyword value and given PCRE* string do not match |
+|   and     | True if expressions on both sides are true                |
+|   or      | True if at least one of two expressions is true           |
+
+\* PCRE = Perl Compatible Regular Expression
+
+
+## Usage examples
+
+### Condition is met always
```
markoa: You see how "Condition is met" repeats every time. Makes reading harder because when you're scanning to find what interests you, you read only headers.

You can simplify it for example:

- Run always
- Run on any branch
- When branch is master
- When branch starts with “df/”
- Run on any tag
- When branch is master or on any tag

etc

DamjanBecirovic: First two would have to be `Skip always` and `Skip on any branch` and that is again tied to use case. 

I will change it to this and we can tweak it when more use cases are available.

- Always true
- On any branch
- When branch is master
- When branch starts with “df/”
- On any tag
- When branch is master or on any tag

markoa:
👍

PR #392, opened 2019-05-05
 added instructions for testflight and hockey app

DmitryBespalov:
- Added new file with instructions.

Filename should be changed before merging.


markoa:
Looks good, thanks!

I'll publish this as two separate pages.

PR #405, opened 2019-05-08
Quotas and limits

shiroyasha:
This document describes various quotas and limits of Semaphore 2.0. It lives in the Reference part of the documentation.

Inspiration documentation that describes quotas on Google Cloud https://cloud.google.com/compute/quotas.

### DIFF
--- a/quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
+++ b/quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
@@ -2,15 +2,14 @@ Semaphore 2.0 enforces usage quotas and limits to protect customers and the
 platform from unforeseen spikes in usage. As your use of Semaphore expands over
 time, your quotas may increase accordingly.
 
-If you expect a notable upcoming increase in usage, you can proactively request
-quota adjustments by sending a request to <support@semaphoreci.com>.
+- [Quotas for Maximum Parallel Running Jobs in an Organization](#quotas-for-maximum-parallel-running-jobs-in-an-organization)
+- [Pipeline and Block Execution Time Limit](#pipeline-execution-time-limit)
+- [Job Log Output Size Limit](#job-log-output-size-limit)
 
-- [Quotas for Maximum Concurrent Running Jobs in an Organization](#quotas-for-maximum-concurrent-running-jobs-in-an-organization)
+## Quotas for Maximum Parallel Running Jobs in an Organization
 
-## Quotas for Maximum Concurrent Running Jobs in an Organization
-
-Every organization has a set of quotas that dictate the maximum number of
-cuncurrently running jobs.
+Every organization has a set of quotas that define the maximum number of
+parallel running jobs.
 
 Default quotas per machine type for a newly created organization:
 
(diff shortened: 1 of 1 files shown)


quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
+Semaphore 2.0 enforces usage quotas and limits to protect customers and the
+platform from unforeseen spikes in usage. As your use of Semaphore expands over
+time, your quotas may increase accordingly.
+
+If you expect a notable upcoming increase in usage, you can proactively request
```
markoa: To me this phrase "a notable upcoming release in usage" doesn't sound like what we usually say during customer calls. We need to address the primary reason for increasing default quotas, which is to support organizations whose needs are simply bigger than this.

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
...
+quota adjustments by sending a request to <support@semaphoreci.com>.
+
+- [Quotas for Maximum Concurrent Running Jobs in an Organization](#quotas-for-maximum-concurrent-running-jobs-in-an-organization)
+- [Pipeline and Block Execution Time Limit](#pipeline-execution-time-limit)
+- [Job Log Output Size Limit](#job-log-output-size-limit)
+
+## Quotas for Maximum Concurrent Running Jobs in an Organization
```
markoa: We don't use the word concurrent/concurrency anywhere on marketing site and app so I recommend saying "parallel" instead.

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
...
+
+Every organization has a set of quotas that dictate the maximum number of
```
markoa: Let's be a bit friendlier than "dictate". :) 

Eg _quotas that define the maximum number of parallel jobs_.

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
...
+| Machine Type  | Default Quota |
+| ------------- | ------------- |
+| e1-standard-2 | 8             |
+| e1-standard-4 | 8             |
+| e1-standard-8 | 8             |
+| a1-standard-4 | 2             |
+
+## Pipeline and Block Execution Time Limit
+
+By default, every pipeline and block is limited to one hour execution time. This
```
markoa: "an hour of execution time"

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
...
+      - sudo apt-get install vim
+```
+
+For detailed explanation, see the [execution_time_limit limit section in the
+Pipeline YAML reference][execution_time_limit_reference].
+
+## Job Log Output Size Limit
+
+Semaphore collects up to 16 megabytes of raw log data from every job in a
+pipeline, which in practice roughly equals 100000 lines of output.
```
markoa: 100,000

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
...
+
+Logs longer than 16 megabytes are trimmed with the following message on the
+bottom:
+
+``` txt
+Content of the log is bigger than 16mb. Log is trimmed.
```
markoa: correct notation is "MB" afaik...

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
```diff
@@ -0,0 +1,81 @@
...
+```
+
+This limit is not adjustable.
+
+For collecting longer textual files, or output from long and verbose process,
+we recommend using a blob store like AWS S3 or Google Cloud Storage. If you have
+a specific use case for longer logs available as part of the job log output,
```
markoa: This last sentence is not compatible with _This limit is not adjustable_.

shiroyasha: Good point. Removed the last sentence.

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
shiroyasha: TIL. I've also updated the backend service to print the correct format.

quotas-and-limits_5cc850ba2c7d3a026fd42e7e.md
shiroyasha: Updated this paragraph, and moved it under the machine quota section.

The new text says:

> If your organization needs are bigger than what is provided with the default
machine quotas, you can request an increase by sending a request to
<support@semaphoreci.com>.

shiroyasha:
@markoa All comments addressed. Let me know if you approve a merge.

PR #412, opened 2019-05-14
Update Rails example to use Compose Style CI

shiroyasha:
Related changes in the demo repository: https://github.com/semaphoreci-demos/semaphore-demo-ruby-rails/pull/3

---

One sentence that is not yet verified is:

> Firefox, Chrome, and Chrome Headless drivers for Capybara work out of the box,
> so you will not need to make any adjustment for browser tests to work on
> Semaphore.

The tests in the example repository are passing, but I haven't verified if they are truly using the Browser to run tests, or only a built in HTML renderers.


markoa:
Closing since we're not going to replace the current approach, but provide a pointer to the Rails example project on Docker in the _Custom CI/CD environment with Docker_ page.

PR #413, opened 2019-05-15
Edit quota section

ivanaur:


billing_5c06ad9604286304a71cfa5b.md
```diff
@@ -9,25 +9,19 @@ the first organization you create on Semaphore 2.0. If needed, an owner of
...
 starting from the moment you created an organization. When you add your credit
 card, you will receive $200 gift credit for your organization that you can use
 during a 3-month period.
 
-On this plan, you can also run 8 jobs
-[per machine](https://docs.semaphoreci.com/article/20-machine-types) in parallel
-by default (total 24 parallel jobs). If you’d like to scale up and run more
-than 8 jobs in parallel, you can easily do that by reaching out to
+Every organization has a set of quotas that define the maximum number of parallel running jobs.
+You can see those numbers [here](https://docs.semaphoreci.com/article/133-quotas-and-limits#quotas-for-maximum-parallel-running-jobs-in-an-organization).
```
markoa: Let's teach users page names they can expect, instead of "here/this page".

Also I think that we should avoid linking directly to a `#page-section` because a change in header title invalidates links. While URLs will be "forever". But I'm not 100% decided yet.

> You can see those numbers on the [_Quotas and limits_](https://docs.semaphoreci.com/article/133-quotas-and-limits) page.

> Refer to [_Machine types_](https://docs.semaphoreci.com/article/20-machine-types) for more information about available hardware that you can use.

ivanaur: Thank you!

PR #418, opened 2019-05-17
OS Image property in agents becomes optional

shiroyasha:
The main motivation for this change is to allow customers to define an agent with containers, without the need to provide the os_image parameter that feels redundant in this setup.

So far I've updated the following articles:

- Pipeline YAML
- Custom CI/CD environment

@markoa do you maybe recall which other pages could be updated?

One other that I kinda wanted to update, but ended up skipping, is the Machine Types page. It could be expanded with default OS image field. I'm 50/50 about this change atm.

### DIFF
--- a/custom-ci-cd-environment-with-docker_5caf3f432c7d3a392f9ceb50.md
+++ b/custom-ci-cd-environment-with-docker_5caf3f432c7d3a392f9ceb50.md
@@ -34,7 +34,6 @@ name: Hello Docker
 agent:
   machine:
     type: e1-standard-2
-    os_image: ubuntu1804
 
   containers:
     - name: main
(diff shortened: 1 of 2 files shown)


shiroyasha:
Update: The backend support is shipped and tested on this repository https://github.com/renderedtext/compose-ci-playground/blob/master/.semaphore/semaphore.yml#L5.

This per-requisite is ✅

markoa:
Great!

AFAIK the only other places where we use `os_image` are example projects. You may update the Rails one, and others are still WIP so I'll make sure they're updated as they get published.

shiroyasha:
Thanks for the review. 🙇 

The YAML for ruby-rails demo project has been modified in this commit https://github.com/semaphoreci-demos/semaphore-demo-ruby-rails/commit/67b9875d20af2e725a60bb3f2eb9eafa8a0dffe2.

PR #421, opened 2019-05-28
Reference: Document block dependencies property

bmarkons:
This PR introduces `dependencies in blocks` section within pipeline YAML reference.

It explains how the property is used along with one example snippet.

Let me know if there is something you would like to add within a reference.

### DIFF
--- a/pipeline-yaml_5b5744d40428631d7a8940a9.md
+++ b/pipeline-yaml_5b5744d40428631d7a8940a9.md
@@ -449,8 +449,10 @@ blocks:
           - echo "output"
 ```
 
-For more examples of common CI/CD workflows check out this [repository](https://github.com/semaphoreci-demos/semaphore-demo-workflows). Here you can find
-specifications for Fan-in/Fan-out, Monorepo and Multi Platform pipelines.
+For more examples of common complex CI/CD workflows, see the
+[semaphore-demo-workflows](https://github.com/semaphoreci-demos/semaphore-demo-workflows)
+repository on GitHub. You will find specifications for fan-in/fan-out,
+monorepo and multi-platform pipelines.
 
 ### task in blocks
 
--- a/pipeline-yaml_5b5744d40428631d7a8940a9.md
+++ b/pipeline-yaml_5b5744d40428631d7a8940a9.md
@@ -374,14 +374,10 @@ displayed in the Semaphore 2.0 UI.
 
 ### dependencies in blocks
 
-The `dependencies` property of a block is used to define the flow of execution
-between blocks within a pipeline.
+When your pipeline is running blocks in parallel,
+you can use the `dependencies` property to define the flow of execution for subsequent blocks.
 
-To specify that block with name `Block C` needs to be run once the
-blocks with names `Block A` and `Block B` finish, the `dependencies` property of
-the block with name `Block C` needs to be populated with corresponding names
-of the blocks which block with name `Block C` depends on.
-In the following example blocks with names `Block A` and `Block B` would be executed in parallel at the
+The following example runs `Block A` and `Block B` in parallel at the
 very beginning of a pipeline.
 
 ``` yaml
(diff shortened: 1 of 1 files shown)


pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -371,10 +372,58 @@ Error: "There are at least two blocks with same name: Build Go project"
...
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+
+blocks:
+  - name: "A"
+    dependencies: []
+    task:
+      jobs:
+      - name: "A"
```
DamjanBecirovic: Maybe these names should be changed to `Job A` etc. so people would not be confused about which name they should put in `dependencies` list .

bmarkons: This is very good notice. I will change it.

Also I consciously decided to name Semaphore entities in a highly generic way, in order not to confuse people with some specific use case who might not be familiar with it. Feedback about this decision is also welcome.

Thanks!

darkofabijan:
@markoa please take a look.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -77,12 +78,12 @@ version: v1.0
 ## name in preface
 
 The `name` property is a Unicode string that assigns a name to a Semaphore
-2.0 project and is optional. However, you should always give descriptive names
-to your Semaphore 2.0 projects.
+2.0 pipeline and is optional. However, you should always give descriptive names
```
markoa: Let's use the opportunity to remove "2.0" from this paragraph.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -371,10 +372,59 @@ Error: "There are at least two blocks with same name: Build Go project"
 Semaphore assigns its own unique names to nameless `blocks` items, which are
 displayed in the Semaphore 2.0 UI.
 
+### dependencies in blocks
+
+The `dependencies` property of a block is used to define the flow of execution
+between blocks within a pipeline.
```
markoa: Prefer active over passive voice. This applies to all writing.

markoa: It seems like I might want to use this when running blocks in parallel, but it's not clear. Can you rephrase?

Assume that the reader is trying to figure out if this is relevant to them from the first paragraph.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -371,10 +372,59 @@ Error: "There are at least two blocks with same name: Build Go project"
...
+
+To specify that block with name `Block C` needs to be run once the
+blocks with names `Block A` and `Block B` finish, the `dependencies` property of
+the block with name `Block C` needs to be populated with corresponding names
+of the blocks which block with name `Block C` depends on.
+In the following example blocks with names `Block A` and `Block B` would be executed in parallel at the
+very beginning of a pipeline.
```
markoa: This is very hard to read, first sentence is in 4 lines.

pipeline-yaml_5b5744d40428631d7a8940a9.md
markoa: There's almost always a way to say the same thing in active voice, which sounds more friendly, without losing the technical rigor.

For example:

_When your pipeline is running blocks in parallel, you can use the `dependencies` property to define the flow of execution for subsequent blocks._

pipeline-yaml_5b5744d40428631d7a8940a9.md
markoa: Please rewrite with less words, taking into account my earlier comment about active voice.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -371,10 +372,59 @@ Error: "There are at least two blocks with same name: Build Go project"
...
+    task:
+      jobs:
+      - name: "Job C"
+        commands:
+          - echo "output"
+```
+
+*Note*: When `dependencies` property is used to define the names of the blocks which certain block
+depends on, property becomes mandatory for all other blocks within a pipeline
+definition. Otherwise the YAML validation error is being thrown.
```
markoa: This is hard to understand. Can you provide an example of what wouldn't work?

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -371,10 +372,59 @@ Error: "There are at least two blocks with same name: Build Go project"
...
+In case `dependencies` property is not used in any of the blocks within a
```
markoa: This is something else, so make it a separate paragraph.

pipeline-yaml_5b5744d40428631d7a8940a9.md
markoa: I think it means that if you write `dependencies` in one block, then you have to write it in each block?

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -371,10 +372,59 @@ Error: "There are at least two blocks with same name: Build Go project"
...
+### dependencies in blocks
+
+The `dependencies` property of a block is used to define the flow of execution
+between blocks within a pipeline.
+
+To specify that block with name `Block C` needs to be run once the
+blocks with names `Block A` and `Block B` finish, the `dependencies` property of
+the block with name `Block C` needs to be populated with corresponding names
+of the blocks which block with name `Block C` depends on.
+In the following example blocks with names `Block A` and `Block B` would be executed in parallel at the
```
markoa: Can you think of an additional example? I worry that people may not understand the full power of the feature from this one example.

pipeline-yaml_5b5744d40428631d7a8940a9.md
bmarkons: Yes. I reduced the number of words and added one example when the specification is invalid.

pipeline-yaml_5b5744d40428631d7a8940a9.md
bmarkons: The whole reference section of this property aims to explain simple mechanism upon one can build even the most complex workflows. In the last sentence, the reader is linked to the [repo with workflows examples](https://github.com/semaphoreci-demos/semaphore-demo-workflows) where the most common are specified.

bmarkons:
@markoa Let me know if you think something more should appear in this section.

PR #429, opened 2019-05-30
Fail-Fast

DamjanBecirovic:
Here are docs for `fail_fast` on pipeline level. 

It is a bit more detailed than most of other topics in reference because `fail_fast` is not that wildly known term and I think it has to be explained a bit. 

Also, once job level `fail_fast` is finished we should add note here that it will be automatically set when you set pipeline level one.

### DIFF
--- a/fail-fast.md
+++ b/fail-fast.md
@@ -1,0 +1,152 @@
+There are two strategies for running CI tests. A developer either wants to get
+a comprehensive report on all failures in a test suite or to get a fast
+feedback from the system and fail everything as soon as one of the jobs fails.
+
+Semaphore runs all the jobs and prepares a comprehensive report for
+your pipeline by default. By setting a Fail-Fast strategy in your pipelines
+you can control this policy and get a fast response for your pipelines.
+
+One out of two Fail-Fast strategies can be used:
+
+- A 'stop' strategy that stops everything in your pipeline as soon as a failure
+  is detected.
+
+- A 'cancel' strategy that stops only the jobs and blocks in your pipeline that
+  haven't yet started. This option is ideal if you don't want to stop an already
+  started test execution.
+
+[Stopping all jobs in a pipeline on the first failure](#stopping-all-jobs-in-a-pipeline-on-first-failure)
+[Stopping all jobs in a pipeline on the first failure on non-master branches](stopping-all-jobs-in-a-pipeline-on-the-first-failure-on-non-master-branches)
+[The difference between Stop and Cancel Fail-Fast strategies](#the-difference-between-stop-and-cancel-fail-fast-strategies)
+[See also](#see-also)
+
+## Stopping all jobs in a pipeline on first failure
+
+To stop all the jobs in a pipeline on the first failure, set a fail-fast stop
+policy in your pipeline YAML files.
+
+``` yaml
+version: v1.0
+name: Tests
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+
+fail_fast:
+  stop:
+    when: "true"  # enable strategy for branches, tags, and pull-requests
+
+blocks:
+  - name: Unit Tests
+    task:
+      jobs:
+      - name: "Unit Tests 1/2"
+        commands:
+          - make test PARTITION=1
+
+      - name: "Unit Tests 2/2"
+        commands:
+          - make test PARTITION=2
+```
+
+If 'Unit Test 1/2' fails in the above pipeline, the 'Unit Test 2/2' will be
+stopped.
+
+## Stopping all jobs in a pipeline on the first failure on non-master branches
+
+Often the fail-fast behavior is ideal for feature branches during development,
+but it is not ideal for the master branch of your project where a comprehensive
+report is better suited.
+
+By changing the `when` condition to `branch != 'master'`, we can set a policy
+that activates fail-fast only for non-master branches.
+
+``` yaml
+version: v1.0
+name: Tests
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+
+fail_fast:
+  stop:
+    when: "branch != 'master'"
+
+blocks:
+  - name: Unit Tests
+    task:
+      jobs:
+      - name: "Unit Tests 1/2"
+        commands:
+          - make test PARTITION=1
+
+      - name: "Unit Tests 2/2"
+        commands:
+          - make test PARTITION=2
+```
+
+## The difference between Stop and Cancel Fail-Fast strategies
+
+Sometimes it is not desirable to stop everything in a pipeline. Instead, we
+would like to allow already running jobs and blocks to finish, but prevent new
+ones from starting.
+
+For these scenarios, the 'cancel' strategy can be configured for your pipelines.
+
+``` yaml
+version: v1.0
+name: Tests
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+
+fail_fast:
+  cancel:
+    when: "branch != 'master'"
+
+blocks:
+  - name: Linter
+    dependencies: []
+    task:
+      jobs:
+      - name: "Linter"
+        commands:
+          - make lint
+
+  - name: Security Scan
+    dependencies: ["Linter"]
+    task:
+      jobs:
+      - name: "Security"
+        commands:
+          - make scan-for-vulnerabilities
+
+  - name: Unit Tests
+    dependencies: []
+    task:
+      jobs:
+      - name: "Unit Tests 1"
+        commands:
+          - make test
+
+  - name: Integration Tests
+    dependencies: ["Unit Tests"]
+    task:
+      jobs:
+      - name: "Unit Tests 1"
+        commands:
+          - make test
+```
+
+In the above scenario, if linting fails, but the unit tests have already started,
+we will not break them in the middle of the execution.
+
+The Security Scan and Integration Tests will be canceled.
+
+# See also
+
+- [Fail-Fast pipeline YAML property](https://docs.semaphoreci.com/article/50-pipeline-yaml#fail\_fast)
+- [Defining 'when' conditions](https://docs.semaphoreci.com/article/142-conditions-reference)
--- a/pipeline-yaml_5b5744d40428631d7a8940a9.md
+++ b/pipeline-yaml_5b5744d40428631d7a8940a9.md
@@ -350,42 +350,26 @@ However, as blocks are currently executed sequentially, the
 
 ## fail_fast
 
-The `fail_fast` property enables you to set the desired policy for the behavior
-of other blocks when one of them fails.
+The `fail_fast` property enables you to set a policy for a pipeline when one of
+its jobs fail.
 
 It can have two subproperties, `stop` and `cancel`.
 
-At least one of them is required.
-
-If both are set, `stop` will be evaluated first.
+At least one of them is required. If both are set, `stop` will be evaluated first.
 
 Both `stop` and `cancel` properties require a condition defined with a `when`,
 following the [Conditions DSL][conditions-reference].
 
-If this condition is fulfilled for a given pipeline execution, an appropriate
-`fail fast` policy will be set for blocks of that pipeline.
+If this condition is fulfilled for a given pipeline execution, the appropriate
+`fail fast` policy is activated.
 
-If all blocks are run sequentially, `fail fast` policy is not necessary because
-if one block fails, the ones in a sequence after it won't be started and the
-whole pipeline will fail immediately.
+If a `stop` policy is set, all running jobs will be stopped and all pending
+ones will be canceled as soon as possible when one job fails.
 
-However, if you have blocks in your pipeline running in parallel (this can be
-achieved by setting [dependencies](#dependencies-in-blocks) property for each
-block), when one block fails, others which do not directly depend on it will
-continue to execute and use time and resources even though the whole pipeline
-can be considered to have failed.
-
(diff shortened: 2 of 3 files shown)


pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,92 @@ However, as blocks are currently executed sequentially, the
...
+
+blocks:
+  - name: Block A
+    dependencies: []
+    task:
+      jobs:
+      - name: Job A
+        commands:
+          - sleep 10
+          - exit 127
```
shiroyasha: `exit` is a bit hard on our system, it shuts down the execution shell we are running our commands in. This prevents apilogue_always and epilogue_on_fail to be executed for example.

Let's not advertise it any way in our docs. `false` is a good alternative.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,92 @@ However, as blocks are currently executed sequentially, the
...
+version: v1.0
+name: Setting fail_fast stop policy
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+fail_fast:
+  stop:
+    when: "branch != 'master'"
+
```
shiroyasha: As this is the only example in our docs, please also give one for `cancel`. Either in this same example, or in a dedicated section.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+other blocks when one of them fails.
+
+If all blocks are run sequentially, this is not necessary because if one block
+fails, the ones in a sequence after it won't be started and whole pipeline will
+fail immediately.
+
+However, if you have blocks in your pipeline running in parallel (this can be
+achieved by setting [dependencies](#dependencies-in-blocks) property for each
+block), when one block fails, others which do not directly depend on it will
+continue to execute and use time and resources even though whole pipeline will
```
markoa: I think we need

> even though the whole pipeline can be considered to have failed.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+eventually fail.
+
+To optimize this, you can choose to set either `cancel` or `stop` fail_fast
```
markoa: s/fast_fail/fast failing

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+
+If `cancel` policy is set, when one block fails the blocks which have not
+yet started executing will be canceled, but already running blocks will be
+allowed to finish executing. This will provide you with more data for debugging
+without occupying any more resources than what was occupied in a first place.
+
+If `stop` policy is set, all running blocks will be stopped and all pending ones
+will be canceled as soon as possible when one block fails.
+
+The `fail_fast` property can have two sub properties, `stop` and `cancel`.
```
markoa: This should appear before you explain the purpose of each.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+
+At least one of them is required and if both are stated, `stop` will be
+evaluated first.
+
+Both `stop` and `cancel` properties have only one required property, and that is
+`when` property which holds condition written in
```
markoa: which defines a condition using the Conditions DSL.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+[Conditions DSL][conditions-reference].
+
+If this condition is fulfilled for given pipeline execution, appropriate
```
markoa: by "given pipeline execution" you mean workflow?

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+`fail_fast` policy will be set for blocks of that pipeline.
+
+### An example for setting fail_fast stop policy
+
+In the following example blocks A and B will start immediately and block C will
+wait for block B to be finished.
+
+After 10 seconds when Block A fails, if pipeline was initiated from non-master
```
markoa: This imo would be easier to read:

> When Block A fails, if the workflow was initiated from a non-master branch, the `stop` fast-failing policy will be applied to:
> - Stop the block B
> - Cancel (ie. not run) the block C

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+        task:
+          jobs:
+          - name: Job C
+            commands:
+              - sleep 60
+```
+
+### An example for setting fail_fast cancel policy
+
+In the following example blocks A and B will start immediately and block C will
```
markoa: In the following configuration, blocks A and B run in parallel. Block C runs after block B is finished.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+wait for block B to be finished.
+
+After 10 seconds when Block A fails, if pipeline was initiated from non-master
```
markoa: When Block A fails, if the workflow was initiated from a non-master branch, the `cancel` fast-failing policy will be applied to:

- Let the block B finish
- Cancel (ie. not run) the block C

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+        dependencies: [Block B]
+        task:
+          jobs:
+          - name: Job C
+            commands:
+              - sleep 60
+```
+
+More examples of conditions for frequent use cases can be found
+[here][when-repo-fail_fast-exemples].
```
markoa: Please no. :) Users need to get all information they need within the docs website.

pipeline-yaml_5b5744d40428631d7a8940a9.md
DamjanBecirovic: We had a meeting with @darkofabijan and @shiroyasha where we decided to go with `fail_fast` as a yaml definition property instead of `fast_fail` or `fast_failing`.

Apparently the `fail fast` is the term most of industry uses and it is what you need to search for to find similar behaviors on other services and what will new users most probably look for.

At same time we found that search for `fast failing` only yields Semaphore Classic docs  as a result and so it might be an improperly coined term or a term which is not intuitive for native English speakers.

The other thing I'm not sure here is should we use `fail_fast` or `fail fast` when talking about general behavior policy instead of yaml definition property? 

I used same form everywhere for consistency but that might not be correct.

pipeline-yaml_5b5744d40428631d7a8940a9.md
DamjanBecirovic: I would rather use pipeline here. 

The conditions are evaluated for a particular pipeline. 

Although all attributes (ie. branch and tag) currently used in conditions are also same for each pipeline in whole workflow, that probably won't always be the case in the future (ie. execution result, values of environment variables etc.).

This is why I think we should stimulate people to always consider conditions in proper context, which is in this case a pipeline.

pipeline-yaml_5b5744d40428631d7a8940a9.md
DamjanBecirovic: I will remove this then. 

Those examples are really useful though, and they could eventually be included in some kind of guide or examples collection page in docs, and we can then add link here.

DamjanBecirovic:
@markoa thanks a lot for review, it really helps.

I addressed most of the requested changes and left a few comments, please take a look again when you have time.

pipeline-yaml_5b5744d40428631d7a8940a9.md
markoa: OK. My comment was mostly about your last point, using a human phrase when describing the behavior instead of the YAML keyword. I'll let you pick the best one.

pipeline-yaml_5b5744d40428631d7a8940a9.md
markoa: 👍

pipeline-yaml_5b5744d40428631d7a8940a9.md
markoa: Can you include them as part of this iteration?

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
 `Building executable` block should finish in 5 hours minus the time it took the
 `Creating Docker Image` block to finish.
 
+## fail_fast
+
+The `fail_fast` property enables you to set a desired policy for behavior of
+other blocks when one of them fails.
+
+It can have two sub properties, `stop` and `cancel`.
```
markoa: subproperties

markoa: subproperties

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+
+At least one of them is required and if both are stated, `stop` will be
```
markoa: Break this in two sentences. Say "set" (simpler) instead of "stated".

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+evaluated first.
+
+Both `stop` and `cancel` properties have only one required property, and that is
```
markoa: We can simplify this:

> Both `stop` and `cancel` properties require a condition defined with a `when`, following the Conditions DSL.

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+`when` property which defines a condition using the
+[Conditions DSL][conditions-reference].
+
+If this condition is fulfilled for given pipeline execution, appropriate
```
markoa: a given

markoa: an appropriate

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+However, if you have blocks in your pipeline running in parallel (this can be
+achieved by setting [dependencies](#dependencies-in-blocks) property for each
+block), when one block fails, others which do not directly depend on it will
+continue to execute and use time and resources even though the whole pipeline
+can be considered to have failed.
+
+To optimize this, you can choose to set either `cancel` or `stop` fail_fast
+policy.
+
+If `cancel` policy is set, when one block fails the blocks which have not
```
markoa: If a (also below)

pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -346,6 +347,139 @@ However, as blocks are currently executed sequentially, the
...
+yet started executing will be canceled, but already running blocks will be
+allowed to finish executing. This will provide you with more data for debugging
+without occupying any more resources than what was occupied in a first place.
```
markoa: The sentence can work with just:

> without using additional resources.

shiroyasha:
With the recent introduction of task-level fail-fast-ing, some changes needed to be introduced.

I've modified the following:

- The Pipeline YAML reference
- Introduced a "Fail-Fast: Stop running tests on first failure" use case documentation page

I still need to "link in" the subtitltes in the document, but I only start doing that when I get a green light about the content of the article.

@DamjanBecirovic @markoa I need your review and constructive criticism 🌹 💌

fail-fast.md
```diff
@@ -0,0 +1,155 @@
...
+version: v1.0
+name: Tests
+agent:
+  machine:
+    type: e1-standard-2
+    os_image: ubuntu1804
+
+fail_fast:
+  stop:
+    when: "branch =~ '.*'"
```
DamjanBecirovic: Maybe it is better to use `when: true` or `when: 'true'` here, otherwise it will not work for tags and pull requests.

markoa:
I'm not able to review this line-by-line in the next few days. I propose that you review the narrative part of the article through a free Grammarly account (there are some obvious grammar issues, like "an fast feedback"). Then review the words our users have used when asking for this feature (in the feature request). Make sure the document speaks that language. When you're done, merge. :)

shiroyasha:
Thanks for greenlighting this route, it unblocks me on this project. 🎉 

I'll make the adjustments and merge during the day.

PR #446, opened 2019-07-02
Create skip-building-some-commits-with-ci-skip

rtgkurel:


skip-building-some-commits-with-ci-skip_5d1b58c604286305cb87ffb6.md
```diff
@@ -0,0 +1,5 @@
+## How to skip building some commits with [ci skip]
```
markoa: Header is not necessary, it will auto-appear based on the title set in Help Scout (we should add this to Readme...).

That + I recommend the Use cases category → merge.

PR #463, opened 2019-07-25
Create draft docs for deploying to AWS EB

csidyel:
I can't add certain stuff from here so I guess I'd need to do it from HS.


markoa:
Published on https://docs.semaphoreci.com/article/150-deployment-to-aws-elasticbeanstalk

Also linked on https://docs.semaphoreci.com/article/123-tutorials-and-example-projects

Thanks!

PR #475, opened 2019-07-31
Add auto-cancel in pipeline yml reference docs

DamjanBecirovic:
Here are docs about the auto-cancel feature in the pipeline yml reference page.

This is a bare minimum for the feature to be considered as done, and it might be a good idea to also add a separate page in the `Use Cases` section. 
If you agree with this, I will add it in a separate PR.


markoa:
I agree on adding a page for _Use cases_. Maybe it can include a plain text "diagram" like:

Let's say your Git commit log looks like this:

```
- commit C <- queued
- commit B <- queued
- commit A <- a workflow / pipeline is currently running for this revision 
```

If you write :

```yaml
cancel: running
```

Then the this will happen:

```
- commit C <- pipeline running
- commit B <- canceled
- commit A <- pipeline stopped & canceled
```

PR #476, opened 2019-08-01
Create project-workflow-trigger-options_5d42dcce042863478675bb99.md

radwo:


markoa:
@radwo is the rollout plan to polish this document, then enable the feature, then publish, or...?

radwo:
@markoa the backend and API part of the feature is already deployed. UI part is also ready, but we are waiting with deployment for this documentation.

So yes this is the plan, to first polish this document and then enable the feature to the public.

Additionally, there are two more changes in docs

Project yaml reference https://github.com/semaphoreci/docs/pull/465
Environment Variables  https://github.com/semaphoreci/docs/pull/464

PR #480, opened 2019-08-06
Add Use Cases page on auto-cancel

DamjanBecirovic:
Here is the separate page on the auto-cancel for Use Cases section.

For page title, I thought of `Auto-cancel pipelines on a new push`, but all other ideas are welcome.

### DIFF
--- a/auto-cancel.md
+++ b/auto-cancel.md
@@ -1,19 +1,20 @@
 The *auto-cancel* strategy defines the behavior of pipelines when they enter the
 queue for execution.
 
-By default, all pipelines will be queued and executed on a first-come, first-served
-basis.
+By default, all pipelines will enter the First-In-First-Out (FIFO) queue and wait
+for their time for execution.
 
-Often this is not the optimal solution, especially during development.
+In some cases, this is not the optimal solution, especially if you consider the
+following scenario.
 
-We all head situations where you push some changes, only to realize that you have
+We all had situations where you push some changes, only to realize that you have
 missed something small, so you push a new revision immediately and then find out
 another mistake...
 
-Before *auto-cancel* strategies the only way to get immediate feedback on the
-latest revision was to manually stop all pipelines from obsolete commits.
+Without *auto-cancel* strategies the only way to get immediate feedback on the
+latest revision is to manually stop all pipelines from obsolete commits.
 
-Now if you set up an auto-cancel strategy this will be done automatically.
+If you set up an auto-cancel strategy this will be done automatically.
 That way you will get faster feedback on revisions that matter while skipping
 all the intermediate ones.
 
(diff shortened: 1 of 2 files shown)


auto-cancel.md
```diff
@@ -0,0 +1,164 @@
+The *auto-cancel* strategy defines the behavior of pipelines when they enter the
+queue for execution.
+
+By default, all pipelines will be queued and executed on a first-come, first-served
+basis.
+
+Often this is not the optimal solution, especially during development.
+
+We all head situations where you push some changes, only to realize that you have
```
shiroyasha: typo, head -> had

auto-cancel.md
```diff
@@ -0,0 +1,164 @@
...
+missed something small, so you push a new revision immediately and then find out
+another mistake...
+
+Before *auto-cancel* strategies the only way to get immediate feedback on the
+latest revision was to manually stop all pipelines from obsolete commits.
+
+Now if you set up an auto-cancel strategy this will be done automatically.
+That way you will get faster feedback on revisions that matter while skipping
+all the intermediate ones.
```
shiroyasha: The temporal structure of `Before *auto-cancel* strategies` and `Now if you set up an auto-cancel` is more appropriate for blogs where you are announcing a new feature.

For documentation, it's better to keep the document time-neutral. Example would be:

Without auto-cancel...

If you set up auto-cancel...

auto-cancel.md
```diff
@@ -0,0 +1,164 @@
+The *auto-cancel* strategy defines the behavior of pipelines when they enter the
+queue for execution.
+
+By default, all pipelines will be queued and executed on a first-come, first-served
+basis.
```
shiroyasha: This is still true with the auto-cancel feature too. The FIFO order is not broken. Auto-cancel only "speeds up" the FIFO processing, and cancels redundant items.

auto-cancel.md
```diff
@@ -0,0 +1,164 @@
...
+
+The execution of pipeline from commit A will be stopped and pipelines initiated
+from commits B and C that were waiting in the queue will be canceled.
+
+This will allow pipeline from newest commit D to start executing and provide you
+with feedback as soon as possible after a push.
+
+## Auto-cancel all pipelines on a new push to a non-master branch
+
+Often the auto-cancel strategy is ideal for feature branches during development.
```
shiroyasha: I would drop the `Ofter the` part of the sentence. It is ok to be prescriptive in use cases.

auto-cancel.md
DamjanBecirovic: The later sentence should be related more with `all pipelines will be queued` part than FIFO, but I agree it is not so clear. I will rework that part.

markoa:
Please post a comment after you do a round of edits, thanks.

DamjanBecirovic:
@markoa edits are done.

auto-cancel.md
```diff
@@ -0,0 +1,165 @@
+The *auto-cancel* strategy defines the behavior of pipelines when they enter the
```
markoa: Maybe rephrase this as "when a queue for execution forms"?

As is now it kind of sounds like you're describing any pipeline workflow in daily life...

PR #483, opened 2019-08-06
pipeline-reference: Add global_job_config_docs

### LINKED ISSUE #447: Global Job Config
We have an undocument very useful feature in our YAMLs — `global_job_config`.

Example of global_job_config that will apply the configuration to all jobs in a pipeline:

```
global_job_config:
  secrets:
    - name: raisins
  env_vars:
    - name: MEANING_OF_LIFE
       value: "42"
```

DamjanBecirovic:
Here are the docs for `global_job_cnofig`, I remembered having it almost defined [here](https://github.com/semaphoreci/docs/issues/447) and that sped up the process.

Closes #447

### DIFF
--- a/pipeline-yaml_5b5744d40428631d7a8940a9.md
+++ b/pipeline-yaml_5b5744d40428631d7a8940a9.md
@@ -596,9 +596,22 @@ global_job_config:
     - name: TEST_ENV_VAR
       value: test_value
 blocks:
-  - task:
+  - name: Linter
+    task:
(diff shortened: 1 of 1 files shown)


markoa:
Looks good, but maybe add one more block to make it more realistic?

DamjanBecirovic:
@markoa can you take a look again and hit approve if everything is ok now?
It can not be merged without the approving review.

PR #520, opened 2019-09-10
Removing Artifacts Management from the command list

mimimalizam:
Current state:
image


mimimalizam:
After the change:
image

PR #523, opened 2019-09-19
Update to include source commands.sh based on customer input.

csidyel:


shiroyasha:
@csidyel it seems that this PR has a conflict with the files on master? Can you resolve them, or maybe eaasier: create a new PR from master?

PR #524, opened 2019-09-20
Added available features docs page

ssamardzic:


shiroyasha:
I've missed this update. @ssamardzic what does this change do in the docs?

ssamardzic:
@shiroyasha 

Filip proposed that we add this to docs a couple of months ago, but then we decide that it's better not to add it, at least not in this form. Can be closed.

PR #541, opened 2019-09-30
Changelog: Week of September 30, 2019

shiroyasha:
Opening a placeholder PR for changes made in the week of September 30, 2019.

PR #545, opened 2019-10-03
Add docs for auto_promote using Conditions DSL

DamjanBecirovic:


markoa:
@shiroyasha ping, we need this to complete the changelog entry for the week. #541

markoa:
Should we also update the [intro to deployment](https://docs.semaphoreci.com/article/67-deploying-with-promotions) in guided tour? If yes @DamjanBecirovic can you replace the current examples with the new syntax?

DamjanBecirovic:
We should eventually change all the examples of the previous syntax in docs with a new one, but maybe for the beginning only that one would be ok.

I'll change this one now, and later open the issue for updating the rest of the examples.

markoa:
Sounds good, thanks.

PR #546, opened 2019-10-03
Status in project YAML

radwo:


sem-reference_5b6968642c7d3a03f89d6bff.md
```diff
@@ -924,6 +925,61 @@ spec:
...
 the project.
 
+### Changing Github status notifications
+
+By default, Semaphore will create status checks only for initial pipeline.
+
+To modify this, edit the status property in repository section in the project
+YAML.
+
+For example, to modify the granularity of statuses from pipeline to block level
```
markoa: I don't understand what block-level means with the YML below. What's different when you set level to block?

My guess is that, for example, if your `semaphore.yml` defines two sequential blocks A and B — then you could say "I want my GH status to be just based on block A". Is that the case?

sem-reference_5b6968642c7d3a03f89d6bff.md
```diff
@@ -924,6 +925,61 @@ spec:
...
+        #
+        # Edit this line to set a new level of granularity
+        #
+        level: "block"
+```
+
+When you save the changes and leave the editor, your changes will be applied to
+the project.
+
+To add statuses from promoted pipelines modify the `pipeline_files` entry and
```
markoa: There should be a subheader above to clearly divide this as a different example.

sem-reference_5b6968642c7d3a03f89d6bff.md
radwo: I rewrite this part, not it should be more clear. Thanks for feedback.

markoa:
Great, I've just rephrased some words here and there.

PR #584, opened 2019-11-13
add nvm to autocache

CommanderK5:


CommanderK5:
Yes, it is a bit long and confusing. I'll move it to a separate point.

markoa:
Btw would be great to support CocoaPods.

CommanderK5:
Noted, will create a feature request for that.

PR #585, opened 2019-11-14
pipeline-ref: Add docs for run property of block

DamjanBecirovic:


pipeline-yaml_5b5744d40428631d7a8940a9.md
```diff
@@ -751,9 +752,11 @@ Its result_reason will be set to `skipped` and other blocks which depend on it
 passing will be started and executed as if this block executed regularly and all
 of its jobs passed.
 
-Examples for frequent use cases can be found [here][when-repo-skip-exemples].
+*Note*: It is not possible to have both `skip` and [run](#run-in-blocks)
+properties defined for the same block since both of them configure the same
+behavior, but in opposite ways.
 
-### Example of Blocks
+Here is the example of a block that is skipped on all branches except on master:
```
shiroyasha: Don't use placeholder mechanics in the docs — "here is" — it is given by the context.

Alternatives:

- An example of...
- The following example...

PR #591, opened 2019-11-18
use-cases: Add 'Building monorepo projects' page

DamjanBecirovic:
Any feedback is welcome.

### DIFF
--- a/building-monorepo-projects_5dd3c0782c7d3a7e9ae41d93.md
+++ b/building-monorepo-projects_5dd3c0782c7d3a7e9ae41d93.md
@@ -1,3 +1,5 @@
+*Note*: The monorepo support is currently in `beta` stage of development.
+
 This guide shows you how to optimize your Semaphore workflow for monorepo
 projects.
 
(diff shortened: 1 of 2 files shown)


use-case-monorepo-project.md
```diff
@@ -0,0 +1,272 @@
+This guide shows you how to optimize your Semaphore workflow for monorepo
+projects.
```
shiroyasha: A gentle two-three paragraph introduction into "what is a monorepo" would be a great start for this guide.

Followed up by, "Semaphore comes out-of-box with support for monorepos".

use-case-monorepo-project.md
```diff
@@ -0,0 +1,272 @@
...
+
+## A basic monorepo project setup
+
+Let's say you have a fairly simple monorepo project that consists of:
+
+- A WEB server application located inside `/web-app/` directory
+- An iOS client application located inside `/ios/` directory
+- A separate docs web page located inside `/docs/` directory
+
+A Semaphore pipeline that is *not* optimized for a monorepo project, in this
```
shiroyasha: It is very easy to miss this `not` word here. 

I would avoid giving a full example of a non-optimized monorepo support. Some customers might copy this first paragraph without reading through the whole document.

building-monorepo-projects_5dd3c0782c7d3a7e9ae41d93.md
```diff
@@ -0,0 +1,230 @@
+This guide shows you how to optimize your Semaphore workflow for monorepo
+projects.
```
shiroyasha: Add a note that monorepo support is beta, and feel free to :shipit:

PR #593, opened 2019-11-20
Choosing between a VM and Docker-based environment

shiroyasha:
Inspired to write this based on some recent questions on support. Hopefully, this page sheds some light on this topic and gives rules of thumb when you use which approach.

### DIFF
--- a/environment-types.md
+++ b/environment-types.md
@@ -1,0 +1,170 @@
+# Environment Types
+
+Semaphore supports two types of environments for running jobs. A Virtual Machine
+(VM) based environment, and a Docker container based environment.
+
+- [Virtual Machine based environments](virtual-machine-based-environments)
+- [Docker based environments](docker-based-environments)
+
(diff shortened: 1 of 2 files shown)


environment-types.md
```diff
@@ -0,0 +1,170 @@
...
+
+    - name: cache
+      image: redis:5.0
+```
+
+Read more about [docker based environments][docker-based].
+
+## When should you use Virtual Machines for your projects?
+
+- **You want to have an up to data testing environment with the latest software
```
radwo: s/data/date/

markoa:
I think that it's useful to have this content. Structure wise, I have no other ideas atm apart from introducing a new page as you have. I'd just consider a title that gives you more reason to open it :) eg. _Choosing between a VM and Docker-based environment_.

shiroyasha:
Thanks for the help @markoa. 

Grammarly and manual spell-check and content-check done. Shipping.

PR #602, opened 2019-11-27
Update notifications-yaml-reference one webhook URL per channel needs to be configured

csidyel:
Coming from https://secure.helpscout.net/conversation/1016722892/23902/


bmarkons:
As far as think, we should advise people not to use channels in the notification specification because notification won't work as expected. The only thing we can do is remove channels from the spec. But we can't do that very easily because some customers are using legacy webhooks.

radwo:
Already merged in different PR.

PR #613, opened 2019-12-10
Add docs for webhook notifications

radwo:


markoa:
It seems like someone more familiar with implementation would be a better reviewer than myself.

I can only recommend comparing our future page with [GitHub's](https://developer.github.com/webhooks/), for intro language / motivational speech skills, structure, etc.

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
...
+``` bash
+$ sem create notification master-pipelines \
+    --projects web \
+    --branches master \
+    --webhook-endpoint [webhook-endpoint]
+```
+
+## Setting up Webhook notifications for multiple projects
+
+When creating the notification, you can specify multiple projects as source,
```
shiroyasha: "When creating a notification"

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
...
+        projects:
+          - example-project
+        results:
+          - failed
+      notify:
+        webhook:
+          endpoint: https://example.org/postreceiver
+```
+
+Note that you can list more than value under `results`.
```
shiroyasha: Somthing is not correct here: "more than value under".

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
...
+Notification**. Add the name of the notification and rules and click on the **Save
+Changes** button.
+
+If you’d like to edit or delete an existing notification, click on the name of
+the notification and at the top right corner, click on **Edit** or **Delete...** button
+and follow the steps from there.
+
+## Notification payload
+
+Payload contains all the information related to a pipeline.
```
shiroyasha: The payload contains all the information related to a pipeline.

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
...
+                    "id": "d4b93a5b-69a5-43e6-ab24-06b095fc49bf"
+                }
+            ]
+        }
+    ]
+}
+```
+
+In this example `revision.pull_request` and `revision.tag` are null because
+payload is related to pipeline from branch build. Information about build type
```
shiroyasha: "build" is a deprecated concept from Semaphore Classic. Refer to executed pipelines as pipeline runs.

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
...
+
+``` bash
+$ sem create notifications example \
+    --branches "master,/hotfix\/.*/" \
+    --webhook-endpoint [webhook-endpoint] \
+```
+
+Regex matches must be wrapped in forward slashes (example: `/.*/`). Specifying a
+branch name without slashes (example: `.*`) would execute a direct equality
+match.
```
shiroyasha: A filter can either be a regex match or a direct match. Specifying a filter with forward slashes (e.g. `/hotfix-.*/`) defines a regex match, while a filter without forward slashes (e.g. `master`) defines a direct match.

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
+Semaphore has integrated Webhook based notifications that are delivered on
+the success or failure of a pipeline.
+
+- [Setting up Webhook notification for a project](#setting-up-webhook-notifications-for-a-project)
```
shiroyasha: Use "webhook" (lower case) in the document. We use uppercase for brand names like Slack, Docker, Github, but not for general concepts like webhooks.

webhook-notifications_5def901604286364bc92af37.md
```diff
@@ -0,0 +1,283 @@
+Semaphore has integrated Webhook based notifications that are delivered on
```
shiroyasha: The word "intgrated" is commonly used to refer to an integration with an outside service. Use:

> Semaphore has webhook based notifications that are delivered ...

notifications-yaml-reference_5be45bb02c7d3a31944dbc98.md
```diff
@@ -118,7 +118,7 @@ The `pipelines` property is optional.
 
 ### notify
 
-The `notify` property holds the `slack` property.
+The `notify` property holds the `slack` and `webhook` property.
```
shiroyasha: The notify property holds the slack and webhook properties.

shiroyasha:
@radwo Overall, the document is 👍 

We have documented the "how" part of webhooks, but it would be great if we could also document the "why" part. In the header section of Github's webhooks, there is a good example of how to do this.

radwo:
@shiroyasha thanks for a great feedback!

radwo:
@shiroyasha I just updated this docs for new format, can you check it ? :)

shiroyasha:
Looks great! :shipit:

PR #614, opened 2019-12-11
Added clarification on the colon syntax for sem create.

daveverwer:
The specifics of using sem create are listed in the command line help but are not specifically mentioned in the documentation. This PR fixes that.


markoa:
Thank you!

PR #651, opened 2020-01-09
docs-creation-instructions

Ivana-Markovic:


Ivana-Markovic:
I have not changed the [Development](https://github.com/semaphoreci/docs/blob/master/README-dev.md#development) part as I am not sure how this changes, maybe you can help.

markoa:
@Ivana-Markovic I'm not in the loop with the new system so I'm no longer the right person to review this. In general the purpose of this file is to let someone from our team who has not created new docs files in the past or forgot but needs to do that now, be able to do that without asking someone else.

PR #655, opened 2020-01-14
Add use case docs for Schedulers

DamjanBecirovic:


docs/use-cases/schedule-a-workflow-run.md
```diff
@@ -0,0 +1,128 @@
+# Schedule a workflow run
+
+This guide shows you how to use Semaphore to set up the scheduled workflow runs
+in your project.
```
shiroyasha: Remove this line. Self-describing sentences give little or no value to the reader.

PR #664, opened 2020-01-15
Semaphore 2.0 Platform update on 15.01.2020 

bogyo210:


dnozay:
I have an issue for the last couple hours.

```
sem-version python 3.7
pip install some-package-that-has-python37-wheels-but-not-python38
``` 
`pip` still tries to install python3.8 stuff which in our case doesn't work.

csidyel:
@dnozay Hi 👋 
Thanks for sharing this!
Could you send me a job or workflow id? It'd be useful for the ongoing investigation.

PR #682, opened 2020-02-13
Add ubuntu1804 spec to heading

bogyo210:


shiroyasha:
Imho, we don't need this. The first sentence is should be clear enough.

Also, with the reststructure of the docs, the context of sem-service should be clear, as this only appears under the Linux VM section.

Screen Shot 2020-02-14 at 12 22 40 PM

markoa:
OK, closing. (We certainly shouldn't need to change the title.)

PR #692, opened 2020-02-27
updating to ssh-add only the created key

csidyel:
when adding `~/.ssh/*` it adds all sorts of stuff like `authorized_keys`, `config`, `known_hosts` and this causes a failure.

### DIFF
--- a/docs/use-cases/using-private-dependencies.md
+++ b/docs/use-cases/using-private-dependencies.md
@@ -62,9 +62,9 @@ blocks:
       prologue:
         commands:
           # Correct premissions since they are too open by default:
-          - chmod 0600 ~/.ssh/*
+          - chmod 0600 ~/.ssh/private-repo
           # Add the key to the ssh agent:
-          - ssh-add ~/.ssh/*
+          - ssh-add ~/.ssh/private-repo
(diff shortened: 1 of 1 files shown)


bogyo210:
The comment and the changes refer to different things.
In ~/.ssh/ all files can have 0400 or 0600, they are private files used for ssh based actions. Shouldn't be a problem chmod 0600 ~/.ssh/*

csidyel:
> The comment and the changes refer to different things.
> In ~/.ssh/ all files can have 0400 or 0600, they are private files used for ssh based actions. Shouldn't be a problem chmod 0600 ~/.ssh/*

@bogyo210 The problem is not `chmod` is `ssh-add`. `ssh-add ~/.ssh/*` will fail because it will attempt to add files are not ssh keys. The reason why I include `chmod` in the change is for the sake of complicity so customers that are not that familiar with this get the picture of what they are doing.
In reality, you don't need to change the permissions of every file in there, just the key you want to add.

shiroyasha:
@csidyel This PR is in conflict with master. Should we close it?

csidyel:
Yes, another change was merged that achieved the same.

PR #694, opened 2020-03-02
Update matrix calculation formula

tombruijn:
It used a minus symbol which is a bit confusing as: 2 - 3 = -1
By using an x (for a multiplication) the formula is accurate.


shiroyasha:
I agree. `-` is probably a typo. Thanks for the PR.

PR #751, opened 2020-04-15
include the git sha in the cache key for the build

mjeffryes:
When we followed this example, we found that reusing the same cache key on every build typically resulted in build results from a previous run being restored instead of deploying the correct changes. Our solution was to include the gitsha in the cache key.


markoa:
@mjeffryes Good suggestion, thanks!

PR #752, opened 2020-04-20
Update semaphore-changelog.md

bogyo210:
Agent port modifyed

### DIFF
--- a/docs/reference/semaphore-changelog.md
+++ b/docs/reference/semaphore-changelog.md
@@ -7,7 +7,7 @@ This page is updated on a weekly basis.
 ### Week of April 20, 2020
 
 - Updates to the ubuntu1804 image:  
-    - TCP port 8000 is no longer occupied by our internal services, they are listening now on TCP/64123. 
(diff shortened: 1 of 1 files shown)


docs/reference/semaphore-changelog.md
```diff
@@ -4,6 +4,12 @@ Thank you for using Semaphore!
 We continuously deploy changes that improve the product for you.
 This page is updated on a weekly basis.
 
+### Week of April 20, 2020
+
+- Updates to the ubuntu1804 image:  
+    - TCP port 8000 is no longer occupied by our internal services, they are listening now on TCP/64123.
```
markoa, line 10: If we're saying that our internal services are listening to port xyz then it means that they're somehow useful to end users, in which case you should mention what they are. :) But if not, just saying that port 8000 is no longer occupied would be OK.

bogyo210: Went with the short version, fewer chances to be misunderstood.

PR #756, opened 2020-04-22
Editing Heroku example to contain info from tickets

csidyel:


shiroyasha:
@csidyel please use an environment variable for this. `HEROKU_APP_NAME`.

Reason: Easier to read the YAML and maintain.

PR #761, opened 2020-04-23
api: Update auth example and add quotes on URLs in curl

DamjanBecirovic:


docs/reference/api-v1alpha.md
```diff
@@ -23,10 +23,10 @@ All API requests require authentication. To authenticate, you need an
 authentication token. You can find your authentication token by visiting your
 [account settings](https://me.semaphoreci.com/account).
 
-Authentication Token is sent as a HTTP header
+Authentication Token should be sent as a HTTP header in all requests:
```
shiroyasha, line 26: change `should be` to `must be`.

shiroyasha: As per https://www.ietf.org/rfc/rfc2119.txt.

PR #766, opened 2020-04-27
Removing extra line.

AleksandarCole:
It looks like this line was there by accident. 
It doesn't seem consistent with the style of other pages (eg. databeses)

### DIFF
--- a/docs/ci-cd-environment/sem-version-managing-language-versions-on-linux.md
+++ b/docs/ci-cd-environment/sem-version-managing-language-versions-on-linux.md
@@ -1,5 +1,3 @@
- # sem-version: Managing Language Versions on Linux
-
 The `sem-version` utility in Linux based virtual machines is used for changing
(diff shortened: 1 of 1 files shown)


docs/ci-cd-environment/sem-version-managing-language-versions-on-linux.md
```diff
@@ -1,5 +1,3 @@
- # sem-version: Managing Language Versions on Linux
-
 The `sem-version` utility in Linux based virtual machines is used for changing
```
bogyo210, line 1: is used `to change` versions of programming languages
 Sounds better in my opinion. Lets ask for a third pair of eyes.

AleksandarCole: But I didn't change that sentence at all. I just removed that one line that is currently at the top of the page.

PR #768, opened 2020-05-04
Update/xcode10

CommanderK5:


docs/ci-cd-environment/macos-xcode-11-image.md
```diff
@@ -1,17 +1,17 @@
-# macOS Mojave Xcode 10 Image
+# macOS Xcode 11 Image
```
AleksandarCole, line 1: Rename to "macOS Catalina Xcode 11 Image" and than we can rename it back when we deprecate Mojave.

PR #776, opened 2020-05-14
Update using-private-dependencies.md

csidyel:
I've reduced the information from the Classic page to the creation of the machine user since it is the most common thing in the tickets. Also, since the creation of the SSH key and secret has been already explained. 
I think we want to push customers to use this approach instead of the other approaches explained in the original document.

### DIFF
--- a/docs/essentials/using-private-dependencies.md
+++ b/docs/essentials/using-private-dependencies.md
@@ -78,5 +78,16 @@ That's all there is to it. You can use the approach to add more deploy
 keys to the `private-repo` secret to cover more projects and reuse the
 secret across other projects.
 
+## Using private Ruby gems
+
+If project's dependencies are hosted on Github, 
+you can enable Semaphore environment to fetch repositories with these steps:
+
+- Create a [machine user][machine-user] with access to repositories.
+- Create a new SSH key pair for the purpose of automation.
+- Add the private SSH key to the project as a secret.
+- Add the public SSH key to the user settings for the machine user.
+
 [secrets]: https://docs.semaphoreci.com/guided-tour/environment-variables-and-secrets/#storing-files-in-secrets
 [deploy keys]: https://developer.github.com/v3/guides/managing-deploy-keys/
+[machine-user]: https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users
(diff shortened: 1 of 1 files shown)


docs/essentials/using-private-dependencies.md
```diff
@@ -78,5 +78,16 @@ That's all there is to it. You can use the approach to add more deploy
...
 
+## Using private Ruby gems
+
+If project's dependencies are hosted on Github, 
+you can enable Semaphore to fetch repositories with these steps:
+
+- Create a [machine user][machine-user] with access to repositories.
+- Create a new SSH key pair for the purpose of automation.
+- Add the private SSH key to the project as a secret.
+- Add the public SSH key to the user settings for the machine user.
```
shiroyasha, line 89: While reviewing this document again, I noticed that we have a dedicated paragraph for this already on line 25.

@csidyel wdyt about rewording that paragraph to be more explicit instead of adding a new section?

---

What we are missing in both the above paragraph and this new section is an answer to "why I would use a machine user?", "when is it better to use deploy keys on a single project vs machine users".

csidyel: @shiroyasha I think it would be more efficient to reword the paragraph to include this machine user guide. 

However, I lack the knowledge to answer the why and when.

shiroyasha: In short, you can have two scenarios:

- If you want to fetch only one gem from one repository, you can attach an SSH key to that repository.
- If you need to fetch multiple gems from multiple repositories, you either need <N> SSH keys (because Github enforces that every repo needs to have a unique keys) or to set up a machine-user with an SSH key and allow him to fetch from those N repositories.

Let me know if you need help. I'm available for chat at any point, or follow up with additional question in this thread

shiroyasha: Notes:

- Github has the terminology "deploy keys". These are just regular plain old SSH keys that are attached to a repository, but the terminology can be confusing.
- If you are attaching the SSH key to the repo they call it "deploy keys", if you attach them to a user they call it SSH keys.

csidyel: @shiroyasha can you check the latest commit? So I can merge

shiroyasha: LGTM. :shipit:

PR #783, opened 2020-05-22
Added note regarding the functionality of caching with keys.

dradulescu:
Added an improvement to our docs to make the usage of caching to a key clearer.


docs/essentials/caching-dependencies-and-directories.md
```diff
@@ -118,6 +118,9 @@ with return code 0.
 In case of insufficient disk space, `cache store` frees disk space by deleting
 the oldest keys.
 
+**Note:** Data from a key is not overwritten by default when running `cache store`.
```
markoa, line 121: ~~Active voice is preferred~~. Prefer active voice:

`cache store` does not overwrite data for an existing key.

you need to delete etc

PR #784, opened 2020-05-26
Added unsupported shell configuration.

dradulescu:


docs/faq/faq.md
```diff
@@ -129,6 +129,18 @@ This will allow the connection to a private remote registry using the self-signe
...
 </details>
 
+<details>
+  <summary id="shell-configuration">Job breaks unexplicably after changing shell configuration</summary>
+  <p>
+Adding
+```
+set -euxo pipefail
+``` 
+to your shell is not supported and will cause the jobs to immediately fail.
```
shiroyasha, line 139: Looks good, but lets be a bit more broad about this.

Setting `set -e` or just `set  -o pipefail` will also break the job.

shiroyasha: Make sure to also mention that sourcing a script with that setting (example: `source ~/my_script` or `. ~/my_script`) will have the same effect on the job.

Let me know if you need help or review of this section.

dradulescu:
@AleksandarCole Yes, the updates are also added. I'll ask Christian as well.

PR #785, opened 2020-05-26
Cidyel/deploy to digital ocean

csidyel:
Coming from this task: https://github.com/renderedtext/tasks/issues/440


shiroyasha:
Nicely written 👏 

Shipit.

markoa:
Please always take a look at the page after publishing.

This one had an extra space next to ` ```yaml` so the page looked like this:

Screen Shot 2020-06-03 at 12 19 08 PM

Should be fixed in https://github.com/semaphoreci/docs/commit/0f093ad7280b5069bfa0da6f71e22963e485fda9

PR #794, opened 2020-06-09
Fix indented to intended

tombruijn:
Indented is indentation of text.
intended is meant for a purpose, in this case "for Semaphore 2.0 users
that create Docker containers in their Semaphore 2.0 projects".

Alternatively you could leave the word out entirely and it wouldn't make
the sentence more difficult to follow.


markoa:
Thanks for reporting, fixed by editing #796.

PR #795, opened 2020-06-10
Changelog update

CommanderK5:


docs/reference/semaphore-changelog.md
```diff
@@ -4,6 +4,9 @@ Thank you for using Semaphore!
 We continuously deploy changes that improve the product for you.
 This page is updated on a weekly basis.
 
+### Week of June 8, 2020
+- macOS Mojave image (macos-mojave-xcode11) has been deprecated.
```
markoa, line 8: This should include a note on what to do if you're using it, or a link to instructions if they already exist elsewhere.

markoa: Also what we mean by "deprecated". Some people use that word to mean something which will be removed after a deadline, some use it for things which have been removed etc.

PR #802, opened 2020-06-18
Add section for Boosters

bmarkons:
Completes https://github.com/renderedtext/tasks/issues/354


bmarkons:
@Ivana-Markovic this should enable you to redirect pages to s2 docs.

bmarkons:
Instead of suggesting knapsack I thought maybe to suggest boosters gem which we are using in our projects.

PR #811, opened 2020-07-06
Add locale with sem-service

csidyel:
Info pulled from https://github.com/renderedtext/tasks/issues/493
Another customer asked for it

### DIFF
--- a/docs/faq/faq.md
+++ b/docs/faq/faq.md
@@ -159,8 +159,8 @@ The database should be available as before, without modifying your application i
 1. Create a Dockerfile with the following:
 ```
 FROM postgres:9.6
-RUN localedef -i ru_RU -c -f UTF-8 -A /usr/share/locale/locale.alias ru_RU.UTF-8
-ENV LANG ru_RU.UTF-8
+RUN localedef -i pt_BR -c -f UTF-8 -A /usr/share/locale/locale.alias pt_BR.UTF-8
+ENV LANG pt_BR.UTF-8
 ```
 2. Rebuild the Postgres image using the locale:
 ```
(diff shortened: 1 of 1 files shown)


docs/faq/faq.md
```diff
@@ -143,6 +143,36 @@ To get the job ID you can use <code>sem get jobs</code> to get the list of all r
...
+  
+1. Create a Dockerfile with the following:
+```
+FROM postgres:9.6
+RUN localedef -i ru_RU -c -f UTF-8 -A /usr/share/locale/locale.alias ru_RU.UTF-8
+ENV LANG ru_RU.UTF-8
+```
+2. Rebuild the Postgres image using the locale:
+```
+docker build - -t postgres:lang < Dockerfile
```
bogyo210, line 167: here 'lang' is not literal, in this specific case it is 'ru'

docs/faq/faq.md
```diff
@@ -143,6 +143,36 @@ To get the job ID you can use <code>sem get jobs</code> to get the list of all r
...
+```
+3. Start the newly created image:
+```
+docker run --rm --net host -d -e POSTGRES_PASSWORD=semaphore --name postgres -v /var/run/postgresql:/var/run/postgresql postgres:lang
```
bogyo210, line 171: here 'lang' at the end should/could be 'ru' if we follow the example above

docs/faq/faq.md
csidyel: updated

docs/faq/faq.md
csidyel: updated

PR #812, opened 2020-07-06
Docs on job priorities

DamjanBecirovic:


docs/essentials/prioritization.md
```diff
@@ -0,0 +1,38 @@
+# Prioritizing jobs
+
+At the end of the day when everyone is committing their work or in cases when there
```
markoa, line 3: Can you please simplify this intro assuming the reader is lazy and would prefer to understand the feature and its value in the first sentence? Rest of the page is OK.

PR #816, opened 2020-07-08
blocks and tasks update

csidyel:


docs/guided-tour/concepts.md
```diff
@@ -1,27 +1,36 @@
...
+sourced from `.semaphore/semaphore.yml`. Additional pipelines triggered via
+promotions are defined in separate files.
 
 ![Semaphore 2.0 concepts diagram](https://storage.googleapis.com/semaphore-public-assets/public/images/semaphoreci2-concepts.png)
 
 ## Blocks & Tasks
 
-_Blocks_ are the heart of a pipeline. Each block has a _task_ that
-defines one or more _jobs_. Jobs define the commands to execute.
+Blocks are the building blocks of a pipeline. Each block has a _task_ defined by
```
csidyel, line 20: @markoa This first sentence looks weird to me wdyt?

markoa: Why specifically?

csidyel: To me, it sounds weird that such a short sentence defines Blocks by being blocks.

markoa: I see, I don't think I've realized the duplication until now. I'll think of something else async.

csidyel: I've thought about this one and probably the best idea that comes to mind is:
`Blocks are the building steps of a pipeline.`
The word step is already used above:
`_Blocks_ define what to do at each step in the pipeline.`
Anything else I can think of changes the sentence almost entirely.

Ivana-Markovic:
I see that this is merged but just wanted to share that in the above examples, blocks can be referred to as units (maybe 🙂): "Blocks are the building **units** of a pipeline."

PR #831, opened 2020-07-27
Queue docs

DamjanBecirovic:


docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,173 @@
...
+  processing: serialized
+```
+
+where the values in brackets are replaced with actual values for a given pipeline.
+
+*Note*: If you want to explicitly define the default queue behaviour in your YAML
+configuration you should omit the `name` and `scope` properties and only configure
+the desired type of `processing`.
+
+## Queue scopes
```
markoa, line 43: Examples should be appear by now if not earlier...

DamjanBecirovic:
@markoa can you take a look again?

markoa:
Rebase pls...

AleksandarCole:
@markoa @DamjanBecirovic are we close to publishing this docs update? I would like to include this feature in changelog for past week.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
+---
+description: See how to assign pipelines to execution queues and run them
+sequentially or how to run them in parallel to save time.
+---
+
+# Pipeline Queues
+
+Pipeline queues allow you to control which pipelines Semaphore must run
+sequentially and which may run in parallel. For example, you may configure
+parallel pipelines on the main branch, while allowing only one deployment to
```
markoa, line 10: For a minute this got me confused by thinking how is this related to parallel blocks. Consider rephrasing, maybe this is a case for using the term workflow instead? Here by "pipelines" I understand you mean "instance of a running pipeline" (aka workflow).

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+production to be running at any given time.
+
+You can configure this behaviour via [queue][queue-reference] property in the
+pipeline's YAML configuration file.
+
+## Default behaviour
+
+By default, the pipelines will be assigned to the same queue and run sequentially  
+if they are initiated from the same branch/tag/pull request with the same
```
markoa, line 19: This sentence is incomplete, lacks predicate.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+configuration, e.g. multiple pushes or multiple promotions.
+
+## Using queues for deployments
+
+You can use queues to prevent parallel runs of deployment pipelines that might
+cause issues in the environment you are deploying to.
```
markoa, line 25: s/are deploying/deploy

Use free Grammarly account to catch this.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+
+If your projects are deployed independently one from another, you can use the
```
markoa, line 27: Passive voice, can be rephrased

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+separate queues for each project by setting the `scope` property of the queue
```
markoa, line 28: Seems like you could just say "by setting `scope: project`".

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+configuration to value **project**.
+
+In the following example, we are configuring the deployment pipeline that should
```
markoa, line 31: Complex sentence, try to rephrase.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+be defined as a [promotion][deploying-with-promotions] to run sequentially in
```
markoa, line 32: Good idea to link to a concept here, but you mention promotions earlier, always link to the first ever mention.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+the queue called `Deployment queue`.
```
markoa, line 33: s/the/a

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+  - name: Deploy
+    task:
+      jobs:
+      - commands:
+          - make deploy
+```
+
+In cases where multiple projects are deployed to the same environment (e.g.
+same Kubernetes cluster) you need the deployment pipelines from those projects to
+queue together. To set that up you should use the queue configuration with the
```
markoa, line 57: To set that up,

same value → equal value (Grammarly will help)

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+      jobs:
+      - commands:
+          - make test
+```
+
+## Conditional queue configurations
+
+In some cases it is useful to have different queue behaviors for same pipeline
+on different branches or maybe the tag-triggered pipelines should be assigned to
+a separate queue etc.
```
markoa, line 121: "etc" is not OK in docs :) Find another way.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+  - name: Unit Tests
+    task:
+      jobs:
+      - commands:
+          - make test
+```
+
+## Conditional queue configurations
+
+In some cases it is useful to have different queue behaviors for same pipeline
```
markoa, line 119: Lacks punctuation, check with Grammarly.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+on different branches or maybe the tag-triggered pipelines should be assigned to
+a separate queue etc.
+
+Solution for these and similar cases is to have multiple queue configurations
+that are only applicable if certain conditions are met.
+
+The Semaphore supports this option trough [conditional queue configurations][cond-queue-defs-reference] as it is illustrated in the example bellow.
```
markoa, line 126: The Semaphore → Semaphore 

Strange to break this into a new paragraph.

passive voice again... You could say (full stop) Here's an example:

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+```
+
+In this example, if the pipeline is triggered from the master branch it will be
+assigned to the queue called `Production` that is defined in the scope of the
+organization and can be shared with other projects.
+
+With that queue configuration, we will achieve that only one production deployment
+is possible at any given time in the whole organization.
+
+On the other hand, if the pipeline is triggered from any git tag it will be
```
markoa, line 179: s/from/by

s/that is/

Use Grammarly.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+assigned to the queue called `Image build queue` that is specific to that project.
+
+That should ensure that images are built and pushed in chronological order.
+
+If neither of the conditions above is true (which is the case for all non-master
```
markoa, line 184: Avoid parenthesis.

docs/essentials/pipeline-queues.md
```diff
@@ -0,0 +1,197 @@
...
+branches and pull request) the third configuration will be used since its
```
markoa, line 185: Punctuation missing.

markoa:
One more thing, this page and most of our docs speak like Workflow Builder doesn’t exist. In practice WB is how we want most people to use Semaphore. I suppose that this feature is currently not available in WB but planned to be added. I suggest also planning to introduce a standard way to note in docs whether a feature is also configurable in WB or only YAML by hand.

docs/essentials/pipeline-queues.md
DamjanBecirovic: Workflow is just a container for pipelines and it has no state on its own, so it is not a good fit here, especially because queuing is configured separately for each pipeline in the workflow. 

I adjusted this slightly, but anything else would IMO just make this intro more complicated and it should be clearer from the rest of the article.

docs/essentials/pipeline-queues.md
DamjanBecirovic: This link needs to be in this section to indicate that the deployment should be implemented as a promotion.

I moved it to the first mention of the `deployment pipelines` at the beginning of this section and that should accomplish both things.

DamjanBecirovic:
@markoa here is the latest batch of requested changes, I also ran it through Grammarly.

I did also run everything through Grammarly free account before pushing previous time, but none of these issues came up.
Instead, some of them were marked as premium issue called "Punctuation in Compound/Complex Sentences", so it might be that they changed something recently to push more towards premium accounts.

Regarding workflow builder, we are working currently on adding those missing features, so this might be solved soon.
If you think that we need that noted in the meantime, I would suggest doing that in the separate PR  for all the affected features.

PR #845, opened 2020-08-11
Minor corrections

simonremotify:
Hi!   I'm Simon Deeley and I've just started as a part-time technical writer for Semaphore.

I've been reading the docs to learn more deeply about Semaphore, and while I was doing that I noticed some opportunities to make minor corrections.

Each correction is in a separate commit.  Hopefully, this will make reviewing and accepting/rejecting each one independently easier.

Hope they're useful!


Ivana-Markovic (approved):
Thank you, this is very welcome and useful coming from a knowledgeable native speaker 🙂🙇‍♀️!

PR #850, opened 2020-08-21
Remove channels from notifications docs

bmarkons:
Being able to provide the list of channels where you want to send your incoming webhook payload is a legacy feature. New way to set this up is through Slack app: https://api.slack.com/legacy/custom-integrations#migrating_to_apps

With that said we can't support the integration with multiple channels.

https://github.com/renderedtext/tasks/issues/906


bmarkons:
We can still use channel override feature with Slack. Closing this.

PR #871, opened 2020-09-11
Update Billing FAQs to reflect UI changes

ivanaur:


csidyel (approved):
Looks good. Let's merge it and see how it looks because it's not the same in GH and the real page.

PR #892, opened 2020-10-19
Fix doc include right variables

csidyel:
The system will expect these other variables when using `image_pull_secrets` for compose style. When running manually it's no problem but for compose style it needs to be certain variables. 
Fixed all variables to match.


AleksandarCole (approved):
Great catch, I did not know this.

PR #912, opened 2020-11-16
Add examples for change_in in conditions reference

csidyel:
Suggestions from: https://semaphore.slack.com/archives/C7BA5TTSM/p1605519625448700?thread_ts=1605518101.446600&cid=C7BA5TTSM

### DIFF
--- a/docs/essentials/building-monorepo-projects.md
+++ b/docs/essentials/building-monorepo-projects.md
@@ -21,44 +21,6 @@ Some advantages of a monorepo approach are:
 
 Semaphore comes with out-of-box support for monorepos.
 
-## Usage examples for change_in
-
-### When a directory changes
-
-```yaml
-blocks:
-  - name: Test WEB server
-    run:
-      when: "change_in('/web-app/')"
-```
-
-### When a file changes
-
-```yaml
-blocks:
-  - name: Unit tests
-    run:
-      when: "change_in('../Gemfile.lock/')"
-```
-
-### When the default branch is main
-
-```yaml
-blocks:
-  - name: Test WEB server
-    run:
-      when: "change_in('/web-app/', {default_branch: 'main'})"
-```
-
-### Exclude the pipeline file
-
-```yaml
-blocks:
-  - name: Test WEB server
-    run:
-      when: "change_in('/web-app/', {pipeline_file: 'ignore'})"
-```
-
 ## An example monorepo project setup
 
 Let's say you have a fairly simple monorepo project that consists of:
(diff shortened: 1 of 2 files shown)


docs/reference/conditions-reference.md
```diff
@@ -283,8 +283,47 @@ The supported map parameters are:
...
+### When a file changes
+
+```yaml
+blocks:
+  - name: Unit tests
+    run:
+      when: "change_in('../Gemfile.lock/')"
+```
+
+### When the default branch is main
```
shiroyasha, line 308: Changing the default branch from master to main

docs/reference/conditions-reference.md
```diff
@@ -283,8 +283,47 @@ The supported map parameters are:
...
+
+```yaml
+blocks:
+  - name: Test WEB server
+    run:
+      when: "change_in('/web-app/', {default_branch: 'main'})"
+```
+
+### Exclude the pipeline file
```
shiroyasha, line 317: Exclude changes in the pipeline file

shiroyasha:
Looks good. 👌 
I left a few micro-change requests.

docs/essentials/building-monorepo-projects.md
```diff
@@ -21,6 +21,44 @@ Some advantages of a monorepo approach are:
 
 Semaphore comes with out-of-box support for monorepos.
 
+## Usage examples for change_in
```
DamjanBecirovic, line 24: This section should be at the end of the document here, before `See also` section.

Also, since here we are talking primarily about monorepo (change_in is just a mechanism in this context), a better option for section title would be something like `Additional examples of monorepo configuration `

docs/essentials/building-monorepo-projects.md
```diff
@@ -21,6 +21,44 @@ Some advantages of a monorepo approach are:
...
+### When the default branch is main
+
+```yaml
+blocks:
+  - name: Test WEB server
+    run:
+      when: "change_in('/web-app/', {default_branch: 'main'})"
+```
+
+### Exclude the pipeline file
```
DamjanBecirovic, line 53: We could add the following below the title (in the same place in the other document also):

_Note_: If you change pipeline file the Semaphore will consider `change_in` as `true`. The following illustrates how to disable this behaviour.

csidyel:
@shiroyasha @DamjanBecirovic I updated the PR based on your comments. Thanks for the feedback.

docs/essentials/building-monorepo-projects.md
```diff
@@ -221,6 +221,47 @@ promotions:
...
+      when: "change_in('/web-app/')"
+```
+
+### When a file changes
+
+```yaml
+blocks:
+  - name: Unit tests
+    run:
+      when: "change_in('../Gemfile.lock/')"
```
DamjanBecirovic, line 241: I just noticed this, it should be:
```
when: "change_in('../Gemfile.lock')"
``` 
notice that there is no `/` at the end since `Gemfile.lock` is a file and not a directory.
Also needs to be changed in the other document.

csidyel: @DamjanBecirovic I tested this last week in my project and it worked with the file with the `/` at the end. I'll update the docs anyway.

csidyel:
@DamjanBecirovic Both docs updated. Let me know if it's good enough to merge it.

docs/essentials/building-monorepo-projects.md
DamjanBecirovic: In that case, it is more of a bug than a feature. Thanks for updating the docs :+1:

DamjanBecirovic (approved):

dradulescu (approved):

PR #928, opened 2020-12-07
Working navigation for latest mkdocs-material

dradulescu:
This should enable navigation for the latest mkdocs-material. A css element also needs to be removed for proper layout.


AleksandarCole (approved):
Will be awesome if it works. Will merge and test in local.

PR #945, opened 2021-01-22
Update permissions

ivanaur:


radwo (changes requested):
Owner and Admins can see everything, and only Members are limited.

radwo (approved):

PR #953, opened 2021-02-09
PHP switching method clarification

dradulescu:
Someone was recently confused about the example, it gives the impression that both commands have to be used. It should be either `sem-version` or `phpbrew`.

### DIFF
--- a/docs/programming-languages/php.md
+++ b/docs/programming-languages/php.md
@@ -52,7 +52,7 @@ additional tools.
 Semaphore uses [phpbrew](https://github.com/phpbrew/phpbrew) to manage
 PHP versions. Any version installable with `phpbrew` is supported on
 Semaphore. Version 7.2 is pre-installed. You can install and switch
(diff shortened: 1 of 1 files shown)


bogyo210 (changes requested):
They are not the same: sem-version downloads a precompiled binary while phpbrew downloads and compiles a fresh instance

dradulescu:
I understand. So `sem-version` should be faster while `phpbrew` can be used to grab versions not already present, right?
Aren't the commands in the example redundant in this case?

bogyo210:
No. Sem-version downloads a precompiled php from our servers, activates it and done. ~10s
Phpbrew downloads the php sources and builds them. ~ 7 minutes

Sem-version uses phpbrew to activate an existing php version on the servers.

dradulescu:
@bogyo210 sorry for the late ping here, but I've added the changes as we discussed. Please let me know if this is clear enough 🙇🏻

bogyo210 (approved):

PR #958, opened 2021-02-19
Fix Operator in conditions example

lukasbischof:
Accordingly to the table below, there should be only one equal sign


shiroyasha:
Good catch 👏 

Thanks for the PR 🙇

PR #961, opened 2021-03-03
Update erlang.md

dradulescu:
The list of available Erlang versions is outdated. Maybe it would make more sense to link directly [towards the versions available in the image](https://docs.semaphoreci.com/ci-cd-environment/ubuntu-18.04-image/#erlang-and-elixir) as an only source of truth, wdyt?


AleksandarCole (approved):

AleksandarCole:
@dradulescu yes please, let's do that. Makes more sense and it's easier to maintain.

AleksandarCole (approved):
lgtm

PR #972, opened 2021-04-06
Monorepo documents

TomFern:
This PR includes:
- The monorepo demo project explanation doc
- An updated monorepo workflow doc with images instead of YAML.

Please note:
- This is the first time I add to this repo, I think I modified mkdocs.yml correctly, but I'm not sure.
- I added images to the monorepo workflow, will they work once published?
- I'll need to correct the paths in the image files before we merge, so it points to the master branch, otherwise they won't work once we delete this branch.


TomFern:
I split this PR into two for easier management:

Examples > Monorepo demo >  #973 
Essentials > Monorepo workflows -> #974

PR #974, opened 2021-04-07
Update Essentials > Monorepo workflow 

TomFern:
This version doesn't have any YAML only screenshots.

Before deleting this branch we need to do 

:%s/tf\/monorepo-workflows\//master\// 

to update the reference to the PNGs


TomFern:
@shiroyasha I've added the YAML back into the doc. We don't have tab support yet in Semaphore documents, but as soon as they are available we'll add them to this doc. Let me know what you think.

PR #984, opened 2021-04-21
Warning for using result in conjunction with when

dradulescu:


AleksandarCole (approved):
lgtm 👍

docs/reference/conditions-reference.md
```diff
@@ -155,6 +155,12 @@ identified with one of the `operators` from above are executed with those values
 
 \* PCRE = Perl Compatible Regular Expression
 
+!!! warning "Keyword usage"
+    The `result` keyword can only be used together with
+    [auto_promotions][auto_promote] but not with `when` conditions since it is
+    redundant and this will lead to unexpected  behavior.
```
DamjanBecirovic, line 161: This `redundant` part is not always true, so I suggest rewriting this to something similar to the following: 
```
Both result and result_reason keywords can only be used in [auto_promote][auto_promote] conditions 
since they are evaluated after the pipeline execution is done and its result is known.
All other when conditions are evaluated during pipeline initialization at which point 
the pipeline execution result is unknown. 
```

DamjanBecirovic:
Also, can you, please, change line 112 in the same file to only mention pipelines? 
GitHub does not allow me to quote it directly here for some reason. 
The line should be:
```
<td> Execution result of a pipeline. Possible values are: passed, stopped, canceled and failed.</td>
```

dradulescu:
I'm truly sorry guys, dropped the ball on this one. I've pushed the required changes, please lmkyt 🙇🏻

PR #987, opened 2021-04-30
Update guided tour > getting started

TomFern:
This is the initial iteration of the first page of the guided tour. Reviwing the file directly from GitHub won't work too well as the doc is using some features only supported by mkdocs. The best way of seeing the result is cloning the repo and running `make server`


markoa:
Superseded by #1028.

PR #989, opened 2021-05-06
Create using-terraform-with-google-cloud.md

csidyel:
Let me know if this should be created as a main or promoted pipeline. Any feedback is welcome.

### DIFF
--- a/docs/examples/using-terraform-with-google-cloud.md
+++ b/docs/examples/using-terraform-with-google-cloud.md
@@ -80,7 +80,6 @@ blocks:
     task:
       secrets:
         - name: terraform-key
-        - name: terraform-token
         - name: gcp
       jobs:
         - name: 'Init Terraform Gcloud'
--- a/docs/examples/using-terraform-with-google-cloud.md
+++ b/docs/examples/using-terraform-with-google-cloud.md
@@ -1,0 +1,130 @@
+---
+description: This guide shows you how to use Semaphore to set up a pipeline using Terraform with Google Cloud.
+---
+
+# Using Terraform with Google Cloud
+
+This guide shows you how to use Semaphore to set up a pipeline
+using Terraform with Google Cloud.
+
+For this guide you will need:
+
+- [A working Semaphore project][create-project] with a basic CI pipeline.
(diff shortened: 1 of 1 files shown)


docs/examples/using-terraform-with-google-cloud.md
```diff
@@ -0,0 +1,130 @@
...
+
+## Store credentials in secrets
+
+- Store the Terraform Personal Access Token
+
+```bash
+sem create secret terraform-token \
+  -e TERRAFORM_TOKEN=<your-personal-token> 
+```  
+
```
bogyo210, line 25: This token was only needed for Digitalocean, the token is created on Digitalocean UI (web) to have a private api-key.
For gcloud only the ssh key and the gcloud.json is needed.
To be able to use the private key on gcloud, the public part must be added to the google-cloud-project 'Metadata/SSh keys'

csidyel: How does it look now? btw is it not necessary to run `ssh-keyscan` or `ssh-add` to make the connection work without any confirmation from our side?

csidyel (commented):
@bogyo210 @AleksandarCole Can I merge this?

docs/examples/using-terraform-with-google-cloud.md
```diff
@@ -0,0 +1,122 @@
...
+### Verify it works
+
+Push a new commit on any branch and open Semaphore to watch a new workflow run.
+If all goes well you'll see a `Passed` box next to your pipeline indicating
+the workflow finished successfully.
+
+### Next steps
+
+Congratulations! You have created a successful pipeline that
+communicates Terraform with Google Cloud.
```
bogyo210, line 105: Maybe a successful pipeline that applies a Terraform recipe to Google Cloud?

bogyo210 (commented):
I think that people will find it interesting to see the actual recipe files, used by terraform. 
I know that was the hardest part to figure out.

csidyel (commented):
@bogyo210 I updated the page as per our sync.

bogyo210 (approved):

PR #1009, opened 2021-05-28
Tweak pipeline_file config in change_in docs

bmarkons:
I had a bit of a hard time understanding what is the pipeline_file config in change_in and how to disable changes in the pipeline YAML file. Since `track` is the default value, I changed the description focus on `ignore`. That will hopefully make instruction on how to change the default behaviour slightly more straightforward.

### DIFF
--- a/docs/reference/conditions-reference.md
+++ b/docs/reference/conditions-reference.md
@@ -251,9 +251,7 @@ The supported map parameters are:
       <td> pipeline_file </td>
       <td>
         Possible values are <b>track</b> and <b>ignore</b>, default is
-        <b>track</b>. Only if track is chosen, the path to the give pipeline file
-        will be automatically added to the paths that are given as a parameters
-        of the change_in function.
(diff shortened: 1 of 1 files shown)


markoa:
wdyt @DamjanBecirovic ?

DamjanBecirovic:
This version is a bit unclear in how using `ignore` affects the result of the change_in.
What do you think about the following @markoa @bmarkons ?

> Any change to the pipeline YAML file will cause `change_in` to be evaluated as `true` with the default `track` value. 
> To avoid this, you should use the `ignore` as a value for `pipeline_file`.

bmarkons:
Sounds much better to me! 👏 :)

AleksandarCole:
@DamjanBecirovic @bmarkons @bmarkons we had a user also confused by this. https://github.com/renderedtext/tasks/issues/2052

DamjanBecirovic:
Closed in favor of #1048

PR #1039, opened 2021-07-15
Update deployment-to-aws-elastic-beanstalk.md

boveus:
We experienced an issue with the `sleep 2` portion of this command.  It was causing the `describe environments` check to get throttled if an EC2 instance took a particularly long time to run.  We bumped it up to a `sleep 5` and it seemed to correct that issue.


markoa:
Thanks @boveus, I'm sure this will help others.

PR #1056, opened 2021-09-09
Fix unclosed link tag on index page

tombruijn:
In commit d95676b27362bcc12f094188d5ac988e337f62d5
this link was changed, but the a-element wasn't closed properly.


markoa:
Thank you Tom!

PR #1062, opened 2021-09-27
New sidebar on essentials page — Preparation to make Tests a key concept on Semaphore 2.0

shiroyasha:
This is the first step in making "Tests" a key concept in our product/documentation.
In this PR, I'm reorganizing the Essentials page to make a proper place under the sun for tests.

The thought process was the following:

**Step 1) Current State in Production**

The list of essentials is too huge. Definitely longer than the average attention span of a reader. 
Also, the list is not ordered in any meaningful way so it is hard to consume.

Screenshot 2021-09-27 at 18 23 48

**Step 2) Essentials organized by topic**

I re-organized the links and grouped them into related categories. The group names emerged after the organization.
Tests are introduced as a key concept in this step. 

Screenshot 2021-09-27 at 18 23 55

**Step 3) Optimizing the visual aspect of the essentials sidebar**

Nested elements in the navigation were visually smushed together. A stronger separation was necessary. I've added some custom CSS to make it nicer.
 
Screenshot 2021-09-27 at 18 27 31

---

The navigation changes kept the links intact. Every page is still under the same URL.

### DIFF
--- a/docs/stylesheets/extra.css
+++ b/docs/stylesheets/extra.css
@@ -56,6 +56,12 @@ header {
     box-shadow: 0 2px 4px 0 rgba(0,0,0,0.16), 0 0 0 1px rgba(0,0,0,0.08);
 }
 
+.md-nav[data-md-level="2"] {
+    margin-left: 2px;
+    border-left: 2px solid black;
+    margin-bottom: 1.5em;
+}
+
 @media only screen and (min-width: 76.25em) {
     .md-tabs {
         background-color: #c8e6c9 !important;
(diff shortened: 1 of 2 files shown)


markoa:
I like this reorg. IMO job matrix + monorepo belong to Optimizations. Private dependencies is an orphan for a reason, it's not a feature but an example.

shiroyasha:
Thank you. I applied this feedback as well, namely:

- Moved Job Matrix and Monorepo to optimizations
- Moved Private dependencies to examples as "Using private dependencies"

This left us with an uncategorized "Debugging with SSH". I moved that to Key Concepts as "Debugging".

End result:
Screenshot 2021-09-27 at 21 10 27

PR #1066, opened 2021-09-28
Reference docs for after_pipeline and environment variables that are injected into after pipeline jobs

shiroyasha:
Grammar is checked with Grammarly.

It reports the overuse of passive voice. I'm keeping it as it is because the rest of the reference is also heavily using passive voice.

### DIFF
--- a/docs/ci-cd-environment/environment-variables.md
+++ b/docs/ci-cd-environment/environment-variables.md
@@ -342,6 +342,9 @@ Example values: `Update Readme.md`
 
 ### Environment Variables injected into after_pipeline jobs
 
+The following environment variables are injected only in
+[after pipeline jobs][after-pipeline-ref]{target="_blank"}.
+
 #### SEMAPHORE\_PIPELINE\_RESULT
 
 The value of the `SEMAPHORE_PIPELINE_RESULT` contains the result of the pipeline.
@@ -443,3 +446,4 @@ This feature adds the following environment variables to every job for a given p
 
 [scheduler]: ../essentials/schedule-a-workflow-run.md
 [setting-env-vars]: ../essentials/environment-variables.md
+[after-pipeline-ref]: /reference/pipeline-yaml-reference/#after_pipeline
--- a/docs/ci-cd-environment/environment-variables.md
+++ b/docs/ci-cd-environment/environment-variables.md
@@ -340,6 +340,64 @@ name of the Pull Request.
 
 Example values: `Update Readme.md`
 
+### Environment Variables injected into after_pipeline jobs
+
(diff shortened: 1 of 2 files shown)


shiroyasha:
For all reviewers:

I'm 90% sure that everything written in the PR is 👌. 
I'm shipping it to prod as I feel that it is already valuable to customers in this form.
 
This doesn't mean that I want to skip the review part of course. 👼 
Feel free to comment and review even after the PR is closed. Feedback is always welcome.

docs/ci-cd-environment/environment-variables.md
```diff
@@ -340,6 +340,67 @@ name of the Pull Request.
...
+
+The value of the `SEMAPHORE_PIPELINE_RUNNING_DURATION` contains the pipeline
+execution time while the jobs were running.
+
+The value is expressed in seconds.
+
+#### SEMAPHORE\_PIPELINE\_CREATED\_AT
+
+The value of the `SEMAPHORE_PIPELINE_CREATED_AT` is the UNIX epoch timestamp when
+the pipeline was created.
```
DamjanBecirovic, line 392: An example of the value here and for the other timestamp env vars could be helpful.
We have example values for all other env vars on this page, so it might be good from a consistency standpoint also.

docs/reference/pipeline-yaml-reference.md
```diff
@@ -1548,6 +1548,50 @@ blocks:
...
+For example, to submit pipeline duration metrics and to publish test results, you
+would define the after pipeline with the following YAML snippet:
+
+``` yaml
+after_pipeline:
+  task:
+    jobs:
+      - name: Submit Metrics
+        commands:
+          - echo "ci.duration:$SEMAPHORE_PIPELINE_TOTAL_DURATION|ms" | nc -w 3 -u statsd.example.com
```
DamjanBecirovic, line 1567: According to the env vars section, this value is in seconds and here it is being sent as milliseconds.
Even though this is just an example, someone can take it literally and end up with wrong values.

docs/reference/pipeline-yaml-reference.md
```diff
@@ -1548,6 +1548,50 @@ blocks:
...
+
+Additionally, in the after pipeline jobs, Semaphore injects environment
+variables that are describing the state, result and duration of the executed
+pipeline.
+
+See [which environment variables][after-pipeline-env-vars] are injected into after pipeline jobs.
+
+### Global jobs config is not applied to after_pipeline jobs
+
+Global job config is not applied to after pipeline tasks. This includes secrets,
```
DamjanBecirovic, line 1591: after pipeline tasks -> after pipeline jobs

PR #1084, opened 2021-10-28
Add non-GA warning on self-hosted agents overview

lucaspin:


markoa:
> Self-hosted agents are now in closed beta! If you would like to join beta and try this feature out, please [share your use case](https://semaphoreci.com/contact) with our team and we will add you to the waiting list and contact you after confirming that the feature suits your requirements. 

Thanks for being careful about this. I think the copy can be improved to sound a bit less defensive, if not hostile. 😅 

Instead of saying closed, exclamation mark!, you will wait, we will evaluate you, and confirm, but maybe not... we could say:

> Self-hosted agents are in closed beta. If you would like to run Semaphore agents in your infrastructure, please [contact us and share your use case](https://semaphoreci.com/contact). Our team will get back to you as soon as possible.

AleksandarCole:
@markoa thank you 🙇 
Updated per your suggestion. Please review and let me know if it sounds better now.

markoa (approved):

PR #1087, opened 2021-11-01
Glossary

shiroyasha:
One challenge I was working on today was to make the terminology for our new Tests tab crystal clear.

During this process, I realized that having this terminology specified in the docs 
will help both us internally, and our customers externally.

@markoa I would like to hear your thoughts about this proposal.

### DIFF
--- a/docs/reference/glossary.md
+++ b/docs/reference/glossary.md
@@ -6,8 +6,8 @@ description: Glossary of commonly used terminology on Semaphore 2.0.
 
 ## Tests
 
-This section describes commonly used terms relating to running, inspecting and
-reporting about tests.
+This section describes commonly used terms relating to running, inspecting,
+and reporting about tests.
 
 #### Test case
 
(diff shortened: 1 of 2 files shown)


docs/reference/glossary.md
```diff
@@ -0,0 +1,95 @@
...
+
+The result of a single test case with one of the following outcomes:
+
+- Passed - The test passed the assertions in the test case
+- Failed - The test failed the assertions in the test case
+- Error - The test case failed to execute. Assertion was not tested.
+- Timeout - The test case failed to execute in preset interval. Assertion was not tested.
+- Skipped - The test case was skipped, and has no result.
+
+#### Test results
```
markoa, line 62: At this point the terminology begins to smell. :) _Test result_ and _Test results_ are too similar as terms for what they refer to.

Perhaps ~~Test result~~ → Test case result && ~~Test results~~ → Test suite results?

markoa: Depending on what you want to do here, there may be changes to the final section of the doc.

shiroyasha: I feel that we would do ourselves a disservice if we didn't cover the term "test results" on this page.
We commonly use this term in various contexts, it must be covered.

I decided to merge these two sections into one unified `test results` term and cover both the singular and plural form.

shiroyasha:
I've covered the feedback points + I did a round of Grammarly on the text.

@markoa looking forward to your review.

markoa (approved):
LGTM!

Now we just have to backfill all older terms. 😅

PR #1099, opened 2021-11-26
Fix links using the hhttps protocol

tombruijn:
I clicked a link and nothing opened. When I opened it in a new tab
Firefox said "The address wasn’t understood".

Looks like the `https` protocol was mistyped. I fixed all occurrences a
quick find all returned.


AleksandarCole:
@tombruijn thank you for both flagging and fixing this one!

PR #1126, opened 2022-02-12
Add AWS support page

lucaspin:


docs/ci-cd-environment/aws-support.md
```diff
@@ -0,0 +1,112 @@
+---
+Description: This guide describes how to set up a fleet of self-hosted agents in AWS
+---
+
+# AWS support
```
markoa, line 5: Can be improved, assuming this is within _Self-hosted agents_ category, I recommend **AWS Auto-Scaling**.

docs/ci-cd-environment/aws-support.md
```diff
@@ -0,0 +1,112 @@
...
+
+!!! beta "Self-hosted agents - closed beta"
+    Self-hosted agents are in closed beta. If you would like to run Semaphore agents in your infrastructure, please [contact us and share your use case](https://semaphoreci.com/contact). Our team will get back to you as soon as possible.
+
+Even though manually starting and stopping agents might serve your purposes, having a dynamically sized pool of agents which automatically resizes itself on high and low load periods allows you to have a more highly available fleet of agents at your disposal.
```
markoa, line 10: The intro starts off in a defensive tone and meanders a bit before landing to the point finally in the third paragraph. The purpose of the intro is to help the reader understand whether they are on the right page. So I recommend cutting the first two paragraphs completely and getting to the point immediately.

markoa (changes requested):
I'm not the best person to review the actual meat of the page. I noticed that the content is more of an essay than a how-to, with many links pointing to our open source repo where "the real" documentation is. I assume you will or have already received feedback on this from someone in the team.

Btw I'd avoid links to like `this#special-flaky-subheader-link-will-become-invalid-if-we-change-one-character` ;)

docs/ci-cd-environment/aws-support.md
lucaspin: Using **AWS Auto-Scaling** might give a wrong impression that this is only if you want autoscaling capabilities, which is not the case. You can use that CDK application even if you only want a statically sized pool of agents

markoa: OK, makes sense.

markoa (approved):

docs/ci-cd-environment/aws-support.md
```diff
@@ -0,0 +1,203 @@
...
+
+## Architecture
+
+<img src="https://raw.githubusercontent.com/semaphoreci/docs/aws-support/public/self-hosted-aws-stack.png" width="100%"></img>
+
+## Networking
+
+By default, the agents will run in your [default AWS VPC][default AWS VPC]. That means that your agents will have a public and a private IPv4 address assigned to them and will run in a public subnet, having its traffic destined to the internet sent to an [internet gateway][internet gateway].
+
+However, since all the communication between agents and the Semaphore API happens unidirectionally (from the agent), the agent instances don't really need to be publicly accessible through the Internet. They only need outbound access to the internet. Therefore, you can also run your agents in a private subnet with no public IPv4 address assigned, and use [NAT devices][nat devices] to send its outbound traffic to the internet.
```
shiroyasha, line 155: `don't really need` feels more like an opinion than a statement. Documentation is better when it is less ambiguous, and prescriptive.  
 
Suggestion: change `don't really need` -> `don't need`.

docs/ci-cd-environment/aws-support.md
```diff
@@ -0,0 +1,203 @@
...
+
+<b>5. Bootstrap the CDK application</b>
+
+The AWS CDK requires a few resources to be around for it to work properly:
+
+```
+npm run bootstrap -- aws://YOUR_AWS_ACCOUNT_ID/YOUR_AWS_REGION
+```
+
+<b>6. Deploy the stack</b>
```
shiroyasha, line 81: Usually, it is better to have subtitles than bolded sentences.

Two benefits:
- Subtitles will appear in the TOC, so you can quickly jump to it.
- Subtitles are linkable, and you can share the link with your team

shiroyasha (approved):
The content is 👌 and consumable.

PR #1135, opened 2022-02-18
Update scheduler docs to cover new improvements

DamjanBecirovic:


AleksandarCole:
@DamjanBecirovic what about the pause/resume scheduler option? Shouldn't that be included in the docs?

AleksandarCole (approved):

PR #1164, opened 2022-03-22
Bump agent AWS support instructions for v0.1.6

lucaspin:


docs/ci-cd-environment/aws-support.md
```diff
@@ -62,7 +64,20 @@ aws ssm put-parameter \
   --key-id <your-kms-key-id>
 ```
 
-Note: if you want to use the default `aws/ssm` KMS, omit the `--key-id` argument.
+Note: you need `kms:Encrypt` permissions on the KMS key you use here. Alternatively, if you want to use the default `aws/ssm` KMS, omit the `--key-id` argument.
```
shiroyasha, line 67: I usually like to avoid the `Note:` at the bottom. Readers only read it once the above command fails.

I would rather suggest:

```
# Store the registration token

You can use the default SSM store to save the registration token, or you can
create a new SSM store for this agent type.

#### Use the default store

aws ssm put-parameter \
  --name <your-ssm-parameter-name> \
  --value "<your-agent-type-registration-token>" \
  --type SecureString


#### (Alternative) Create a new SSM store and store it there as a parameter

TODO: Document how to create one

and then:

aws ssm put-parameter \
  --name <your-ssm-parameter-name> \
  --value "<your-agent-type-registration-token>" \
  --type SecureString
  --key-id <ID>

```

docs/ci-cd-environment/aws-support.md
```diff
@@ -92,19 +105,23 @@ Note: if your key was encrypted using the default `aws/ssm` KMS key, `SEMAPHORE_
 
 ### 5. Bootstrap the CDK application
 
-The AWS CDK requires a few resources to be on hand for it to work properly:
+Using the ARN of the execution policy created in step 3, bootstrap the CDK application:
```
shiroyasha, line 108: Is there a command in `aws` cli to display this arn?

lucaspin: Unfortunately, no.

There is a [`get-policy`](https://awscli.amazonaws.com/v2/documentation/api/latest/reference/iam/get-policy.html), but that one takes an ARN 😢

shiroyasha: 😢

PR #1192, opened 2022-05-25
added note - user restriction when adding agent

mateuscap:


docs/ci-cd-environment/self-hosted-agent-types.md
```diff
@@ -15,6 +15,8 @@ Before running and using self-hosted agents in your pipelines, you need to regis
 3. Add a name and click `Looks good. Register`
 4. Your agent type should be created. Follow the instructions to install the agent on your operating system of choice. The instructions are also available [here][installing-agents].
 
+**Note:** only users with the Admin permission level can add a self-hosted agent type
```
lucaspin, line 18: It might be better to use the `!!! info` cards we usually use for these notes. For example:

```
!!! info "Required permissions to add a new self-hosted agent type"
    Only users with the **Admin** permission level can add a new self-hosted agent type.
```

will render like this:

Screen Shot 2022-05-25 at 16 55 16

lucaspin (approved):

csidyel (approved):

PR #1297, opened 2023-01-16
changelog-january-16

d-stefanovic:


mateuscap (approved):

bogyo210 (changes requested):
Lines should start with capital or small letter, but not both

PR #1362, opened 2023-10-05
Tasks: public documentation

mattrym:
Original documentation for Schedulers was moved to a different file (with `legacy` suffix).


Charismatron (approved):
Copy edited.

PR #1375, opened 2023-12-01
Cypress Integration Documentation

bostjan-cigan-rt:
Added section for integrations in Semaphore documentation.

First integration includes Cypress.

@csidyel Can you please check if any optimizations could be made in the code itself (if I have missed something, like caching)?


docs/examples/integrations/cypress.md
```diff
@@ -0,0 +1,119 @@
...
+
+![Adding command to execute tests](images/cypress-block-2.png)
+
+Now click on the *Prologue* section and insert the following snippet:
+
+``` bash
+nvm install 20
+npm install -g npm
+checkout
+cache restore npm-$SEMAPHORE_GIT_BRANCH-$(checksum package-lock.json)-$(checksum .semaphore/semaphoreyml)
```
csidyel, line 90: @bostjan-cigan-rt Is there a specific reason why you suggest creating the cache key with the checksum of the semaphore.yml file? 

Having a new cache key when the `package-lock.json` or `yarn.lock` changes is good but it might not be a good idea to create new keys when the semaphore.yml changes too.  I'd remove that from the key.

bostjan-cigan-rt: This was basically in the Cypress documentation.

I guess their use case was if anything changed in the flow itself (i.e. you added a new job or something similar) you should rebuild things?

This is their example for parallel (not sure why they use this as a cache key though):

https://github.com/cypress-io/cypress-example-kitchensink/blob/master/.semaphore/semaphore.yml

If you think it's weird, we can remove it. Don't know which scenario this falls in (having a key based on the sempahoreyml checksum as well).

PR #1381, opened 2023-12-19
BrowserStack integration

bostjan-cigan-rt:
- Added BrowserStack subsection
- Added docs for BrowserStack automate with video

### DIFF
--- a/docs/examples/integrations/browserstack.md
+++ b/docs/examples/integrations/browserstack.md
@@ -1,0 +1,66 @@
+---
+Description: This guide shows you how to use BrowserStack Automate in your Semaphore workflows.
+---
+
+# BrowserStack
+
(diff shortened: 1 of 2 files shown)


AleksandarCole (commented):
@bostjan-cigan-rt The video is great 👏 

Consider expanding the docs so that they explain all key points from the video. 
The reader should be able to set this up just based on the docs page alone. 

We don't have to go into too much depth, but some simple step-by-step would be good.

bostjan-cigan-rt:
Those steps are already in the BrowserStack documentation. That's why we link to it in our docs. The video is basically this part of the BrowserStack docs:

https://www.browserstack.com/docs/automate/selenium/semaphore#set-up-a-pipeline-in-semaphore

Wanted to keep it that way so that BrowserStack needs to maintain the docs, otherwise we need to update them on both ends if anything changes.

bostjan-cigan-rt:
Maybe we can just list the steps in the video in our docs and link to further expanded docs on BrowserStack side?

PR #1398, opened 2024-02-02
Adding dynamically editing ppl file

csidyel:
@DamjanBecirovic I think it would be to add a simple IF here. Something I was thinking was:

Reduce parallelism from 10 to 2 IF change_in detects changes in lib (since I already have that in the ppl file). Would you be able to help me out with the change_in logic at a very simple level just to illustrate the point?

### DIFF
--- a/docs/security/configuring-pre-flight-checks.md
+++ b/docs/security/configuring-pre-flight-checks.md
@@ -129,3 +129,17 @@ if is_promotion; then if user_is_allowed; then echo "Promotion allowed."; else f
 The list of users might come from other sources, for example [secrets](/essentials/using-secrets)
 or cloned repository. It is also possible to extend the functionality
 to restrict promotions from specific repositories branches. 
+
+### Dynamically editing your pipeline files
+
+During the initialization job, two important variables are exported:
+
+- `INPUT_FILE`: This variable points to the unmodified pipeline file, in this case, `semaphore.yml`.  
+- `OUTPUT_FILE`: This variable refers to a copy of the original pipeline file (`INPUT_FILE`).
+Ultimately, Semaphore uses this copied file to execute the pipeline.
+
+By modifying the `OUTPUT_FILE`, it's possible to change the pipeline configuration for that particular run.
+
+In the following example, we'll modify the number of parallel jobs in a block from 10 to 2:
+
+`sed -i 's/parallelism: 10/parallelism: 2/' $OUTPUT_FILE`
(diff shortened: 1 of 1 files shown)


docs/security/configuring-pre-flight-checks.md
```diff
@@ -129,3 +129,17 @@ if is_promotion; then if user_is_allowed; then echo "Promotion allowed."; else f
...
+
+- `INPUT_FILE`: This variable points to the unmodified pipeline file, in this case, `semaphore.yml`.  
+- `OUTPUT_FILE`: This variable refers to a copy of the original pipeline file (`INPUT_FILE`).
+Ultimately, Semaphore uses this copied file to execute the pipeline.
+
+By modifying the `OUTPUT_FILE`, it's possible to change the pipeline configuration for that particular run.
+
+In the following example, we'll modify the number of parallel jobs in a block from 10 to 2:
+
+`sed -i 's/parallelism: 10/parallelism: 2/' $OUTPUT_FILE`
```
DamjanBecirovic, line 145: In the following example, we'll modify the number of parallel jobs in a block from 2 to 10 only if the particular push that triggered the workflow contains changes in the `my_lib` directory: 

```bash
my_lib_dir_changed () { spc list-diff --default-branch main | grep -q "^my_lib/"; }
increase_parallelism () { `sed -i 's/parallelism: 2/parallelism: 10/' $OUTPUT_FILE`; }

if my_lib_dir_changed; then increase_parallelism; fi
```

docs/security/configuring-pre-flight-checks.md
```diff
@@ -129,3 +129,17 @@ if is_promotion; then if user_is_allowed; then echo "Promotion allowed."; else f
...
 or cloned repository. It is also possible to extend the functionality
 to restrict promotions from specific repositories branches. 
+
+### Dynamically editing your pipeline files
+
+During the initialization job, two important variables are exported:
+
+- `INPUT_FILE`: This variable points to the unmodified pipeline file, in this case, `semaphore.yml`.  
+- `OUTPUT_FILE`: This variable refers to a copy of the original pipeline file (`INPUT_FILE`).
+Ultimately, Semaphore uses this copied file to execute the pipeline.
```
DamjanBecirovic, line 139: The output file is not just a simple copy of the input file, but actually, a compiled input file that can be used to fully define a pipeline. So, let's change these two lines to something like this:

- `OUTPUT_FILE`: This variable refers to a full pipeline file compiled from the original file (`INPUT_FILE`) during the [pipeline initialization](https://docs.semaphoreci.com/reference/pipeline-initialization/).
Ultimately, Semaphore uses this compiled file to define the pipeline that will be executed.

docs/security/configuring-pre-flight-checks.md
csidyel: I updated the text. Please, check it out now.

DamjanBecirovic (approved):

PR #1428, opened 2024-04-16
adding reference to the insights page 

mateuscap:


radwo (changes requested):
These are two different features.

PR #1464, opened 2024-08-16
FAQ: Adding information about bash -e and similar

tbogosavljevic:
Some commands change the behavior of this feature in Semaphore. We have received 2 tickets about this, and I believe we should add a small disclaimer to avoid this in the future.


diegoguerrad (approved):
@tbogosavljevic leaving the final ok for Mateus since it's tech :)

mateuscap (approved):

